PR #27, opened 2017-06-16
Replace quickstart example with something that works

anguslees:
Update to ksonnet-lib beta.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,25 +37,31 @@ avoid an immediate `Killed: 9`.
 ## Quickstart
 
 ```console
-# This example uses ksonnet-lib
-% git clone https://github.com/ksonnet/ksonnet-lib.git
+# Fetch guestbook example
+% git clone https://github.com/ksonnet/kubecfg.git
+% cd kubecfg
 
-# Set kubecfg/jsonnet library search path.  Can also use `-J` args everywhere.
-% export KUBECFG_JPATH=$PWD/ksonnet-lib
-
-# Hello-world ksonnet-lib example
-% cd ksonnet-lib/examples/readme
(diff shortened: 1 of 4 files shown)


anguslees:
Assumes https://github.com/ksonnet/ksonnet-lib/pull/30 ...

anguslees:
Replaced with a self-contained guestbook example.

sebgoa:
I like to make the frontend service a nodeport, that way it works out of the box on minikube.

README.md
```diff
@@ -37,25 +37,31 @@ avoid an immediate `Killed: 9`.
 ## Quickstart
 
 ```console
-# This example uses ksonnet-lib
-% git clone https://github.com/ksonnet/ksonnet-lib.git
+# Fetch guestbook example
+% git clone https://github.com/ksonnet/kubecfg.git
+% cd kubecfg
```
sebgoa: There is a bit of a flow issue in this quickstart IMHO. maybe add:

```
make kubecfg
```

sebgoa: If you install from release binaries this will work fine, but then you clone the entire repo to get the example jsonnet and the lib.

if you install from source with `go get` then this quickstart makes you clone again...

anguslees:
> I like to make the frontend service a nodeport, that way it works out of the box on minikube.

Good call.  I made it a LoadBalancer so it works just about everywhere out of the box.

README.md
anguslees: Hrm.  I really wanted to just give people the flavour of using the tool here, without losing it in all the setup details.   I think I'll just remove the git clone part entirely and assume that people can work it out if they _actually_ want to run the example (rather than just look at it).

Does that flow better?

PR #28, opened 2017-06-16
Add a flag for extvars

GauntletWizard:
Extvars would be really useful for my usage, though I'm working with workarounds. Not certain this is completely idiomatic, let me know if there's changes.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	goflag "flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -76,6 +80,18 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
...
 	}
 
+	extvars, err := flags.GetStringSlice(flagExtVar)
+	if err != nil {
+		return nil, err
+	}
+	for _, extvar := range extvars {
+		kv := strings.SplitN(extvar, "=", 2)
+		if len(kv) != 2 {
+			return nil, errors.New("Failed to parse extvar: missing '=' in " + extvar)
```
anguslees: This works, but if you're using a generated error string you usually use `fmt.Errorf` for the more powerful formatting abilities:
```go
  return nil, fmt.Errorf("Failed to parse extvar: missing '=' in %s", extvar)
```

PR #29, opened 2017-06-21
Add native convenience functions to parse JSON and YAML

anguslees:


anguslees:
Note that the tests crash on golang 1.7/osx, but work on golang 1.8/osx.  This change is effectively saying we don't support golang 1.7/osx (for now).

PR #31, opened 2017-06-23
add kubecfg check subcommand

ngtuna:
It's still in progress.

- [x] add check subcommand
- [x] add test
- [x] update vendor 

I am getting trouble with `govendor` in order to fetch `k8s.io/kubernetes/pkg/api/validation` with k8s v1.5 (or v1.6). It keeps saying `cannot find package "k8s.io/api/core/v1"` all the time. Could you help at this point @anguslees as you're more familiar with `govendor` than me?


anguslees:
> I am getting trouble with govendor in order to fetch k8s.io/kubernetes/pkg/api/validation with k8s v1.5 (or v1.6). It keeps saying cannot find package "k8s.io/api/core/v1" all the time.

`k8s.io/api/core/v1` didn't exist in k8s v1.5: https://github.com/kubernetes/kubernetes/tree/v1.5.7/api

I expect that whatever depends on this package also needs to be downgraded.

cmd/check.go
```diff
@@ -18,22 +19,12 @@ var checkCmd = &cobra.Command{
...
-			if groupVersion == "v1" {
-				prefix = "api"
-			}
 
 			// Download schemaData for obj
-			schemaData, err := disco.RESTClient().Get().
-				AbsPath("/swaggerapi", prefix, groupVersion).
-				Do().
-				Raw()
+			schemaData, err := downloadSchema(groupVersion)
```
anguslees: It looks like client-go release-3.0 (which kubecfg is now using) has a `disco.SwaggerSchema(groupVersion)` method.  Can we use that?

cmd/check_test.go
```diff
@@ -0,0 +1,100 @@
...
+        "port": 9092
+      }
+    ],
+    "selector": {
+      "app": "kafka"
+    }
+  }
+}`)
+
+	if err := json.Unmarshal(data, &obj.Object); err != nil {
```
anguslees: Since we don't care about testing the actual deserialisation here, an easier alternative would be to just construct a `runtime.Unstructured` object directly - rather than string->`json.Unmarshal`.   (See `sort_test.go` for an example)

cmd/check.go
```diff
@@ -55,6 +46,26 @@ var checkCmd = &cobra.Command{
...
 	},
 }
+
+func downloadSchema(groupVersion string) ([]byte, error) {
+	prefix := "apis"
+	if groupVersion == "v1" {
+		prefix = "api"
+	}
+
+	_, disco, err := restClientPool(nil)
```
anguslees: This makes this function (and anything that calls it) untestable.  You've embedded the fact that this function in turn requires a client pool, and the client config that gets used, etc, etc - making it impossible to replace any of these dependencies to (eg) make an offline test.

The `downloadSchema()` function needs to be changed to be passed the client to use to fetch the schema - and then the caller passes in the specific client to use.  (But see comment above, we can probably just replace this function with `disco.SwaggerSchema(groupVersion)`)

cmd/check.go
```diff
@@ -0,0 +1,60 @@
...
+			if err != nil {
+				return err
+			}
+
+			// Validate obj
+			objData, err := json.Marshal(obj)
+			if err != nil {
+				return err
+			}
+			err = schema.ValidateBytes(objData)
```
anguslees: This is the only actual function we want from `k8s.io/kubernetes`, and it turns out (see other commit in this PR) that that's a *much* bigger dependency than I had expected :(

I think we should just copy the small-ish bit of code in `schema.go` (`ValidateObject` and the functions it calls) and rebase them onto the apimachinery alternatives of the various object types.

anguslees:
Updated in / superseded by #77

PR #32, opened 2017-06-24
fix README

sebgoa:
the example in ksonnet-lib is in a weird `readme` directory, this fixes the basic readme ....

we should have an independent example, so that we don' t have to clone ksonnet-lib...


anguslees:
I fixed this in https://github.com/ksonnet/kubecfg/pull/27 also, but that got stuck behind a ksonnet-lib PR. Thanks for the update.

PR #38, opened 2017-06-27
Update jsonnet-cgo to 04f8990

### LINKED ISSUE #37: RUNTIME ERROR: Field does not exist: selector
This is probably user error, but...

With the following simple testcase:

```jsonnet
local k = import "ksonnet.beta.1/k.libsonnet";

k.core.v1.service.default("frontend") +
  k.core.v1.service.mixin.spec.selector({name: "frontend"})
```

I get this error:
```
$ kubecfg -J production/ksonnet-lib show ./test.jsonnet 
Error: Error reading ./test.jsonnet: RUNTIME ERROR: Field does not exist: selector
		object <mixin>
	production/ksonnet-lib/ksonnet.beta.1/core.v1.libsonnet:3379:31-44	object <anonymous>
	During manifestation	
```

However, it works with plain jsonnet:
```
$ jsonnet -J production/ksonnet-lib ./test.jsonnet 
{
   "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
      "name": "frontend",
      "namespace": "default"
   },
   "spec": {
      "selector": {
         "name": "frontend"
      }
   },
   "status": { }
}
```

The problem seems to stem from the fact that service objects created with `default` have empty `spec` objects, and the `+:` operation is expecting it?  if I change the input to be:

```
local k = import "ksonnet.beta.1/k.libsonnet";

k.core.v1.service.default("frontend") +
  {
    spec+: k.core.v1.serviceSpec.default() +
      k.core.v1.serviceSpec.selector({name: "frontend"})
  }
```

It works.  Should the `default()` function for service be calling the `default()` function for serviceSpec, instead of initialising it to an empty object?

tomwilkie:
Fixes #37

### DIFF
--- a/vendor/github.com/strickyak/jsonnet_cgo/ast.h
+++ b/vendor/github.com/strickyak/jsonnet_cgo/ast.h
@@ -44,11 +44,12 @@ enum ASTType {
     AST_IMPORT,
     AST_IMPORTSTR,
     AST_INDEX,
-    AST_LOCAL,
+    AST_IN_SUPER,
     AST_LITERAL_BOOLEAN,
     AST_LITERAL_NULL,
     AST_LITERAL_NUMBER,
     AST_LITERAL_STRING,
+    AST_LOCAL,
     AST_OBJECT,
     AST_OBJECT_COMPREHENSION,
     AST_OBJECT_COMPREHENSION_SIMPLE,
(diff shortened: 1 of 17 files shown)


anguslees:
Thanks!

PR #39, opened 2017-06-28
Switch logging from glog to logrus

### LINKED ISSUE #34: Improve `stdout` logs
Right now an update shows:

```
$ ./kubecfg update examples/guestbook.jsonnet 
I0627 10:08:39.214140    7694 update.go:57] Updating Service/redis-slave
I0627 10:08:39.277885    7694 update.go:70]  Creating non-existent Service/redis-slave
I0627 10:08:39.296423    7694 update.go:57] Updating Service/frontend
...
```

Probably need to clean up a bit the output and make it more user friendly.

anguslees:
Improve interactive output experience by switching to a logging
library (logrus) that allows customising the output format.

This also removes the glog command line flags.  Replaced with a
simpler `--verbose`/`-v` option:
- quiet by default (warnings and errors only)
- `-v` for progress info
- `-vv` for debug

Fixes #34

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,15 +5,17 @@ import (
 	"encoding/json"
 	goflag "flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
 
-	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	jsonnet "github.com/strickyak/jsonnet_cgo"
+	"golang.org/x/crypto/ssh/terminal"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/pkg/api/unversioned"
@@ -24,6 +26,7 @@ import (
 )
 
 const (
+	flagVerbose    = "verbose"
 	flagJpath      = "jpath"
 	flagExtVar     = "ext-str"
 	flagExtVarFile = "ext-str-file"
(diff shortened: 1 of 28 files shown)


anguslees:
Example output:
output

Errors look similar, but have a red "FATA" prefix:
error

The above is probably not what I would have chosen in a perfect world, but not too bad.  The output is written to stderr, which again is probably ok.

Without a tty, the output is not ok (imo):
```
 % ./kubecfg update -v examples/guestbook.jsonnet 2>&1 | cat
level=fatal msg="Error reading examples/guestbook.jsonnet: RUNTIME ERROR: Couldn't open import "example.libsonnet": No match locally or in the Jsonnet library paths.
        examples/guestbook.jsonnet:16:17-42     thunk <example>
        examples/guestbook.jsonnet:19:24-30     object <anonymous>
        During manifestation
" 
```

So .. assuming I can fix the non-tty case to look basically like the tty version without colours - are we ok with this?

I had dreams of adding a basic progress bar to `update` (particularly once slow garbage collection is added) so this may change again in the future.  I tried just now, and it seems the 3 "popular" golang progress bar libraries don't properly support mixing line-based output and the progress bar (only one claims to support it, and it fails a quick test).

cmd/root.go
```diff
@@ -59,10 +61,34 @@ var RootCmd = &cobra.Command{
 	SilenceUsage:  true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		goflag.CommandLine.Parse([]string{})
-		glog.CopyStandardLogTo("INFO")
+
+		log.SetOutput(cmd.OutOrStderr())
+
+		logFmt := log.TextFormatter{
+			DisableTimestamp: true,
```
sebgoa: I think I would have kept the timestamp

sebgoa:
I don't think you need a progress bar, or if you do it is super low priority.

I am not shocked by the format of the FATAL errors, not sure how you can improve this.

But I would have kept the timestamp in verbose mode.

cmd/root.go
anguslees: If I keep the timestamp, then I'm basically silent (except for actual errors), or looking like the old output.  I thought the original #34 issue was all about removing the timestamp "noise" :/  What would you _like_ the output to look like?

(It's unusual to have timestamps on ephemeral command line tool output, since you're probably looking at the messages as they arrive - unlike messages captured in a log file somewhere)

sebgoa: ok that's fine. The format of the timestamp was just strange.

anguslees:
ok, I'm happy with this version.  Output is similar to screenshots above, but the non-tty version just omits the initial coloured word entirely (a predictable change).  @sebgoa?

sebgoa:
ok, that's fine with me

PR #40, opened 2017-06-28
Rename --extVar to --ext-str

anguslees:
For consistency with other options (hyphenated, not camel case), and
the upstream `jsonnet` tool.


anguslees:
Fyi @GauntletWizard and PR #28

GauntletWizard:
Thanks for the heads up; I whiffed when writing. Thankfully, I'm only using the -V form.

PR #44, opened 2017-06-29
More complete implementation of extvars, topvars

GauntletWizard:
Still needs ext-code, tla-code, etc.


GauntletWizard:
(I wrote 41c5f01 and totally thought it didn't work, because I was looking at go-jsonnet rather than jsonnet_cgo. Happy surprise when it compiled)

cmd/root.go
```diff
@@ -90,10 +97,68 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 	}
 	for _, extvar := range extvars {
 		kv := strings.SplitN(extvar, "=", 2)
+		switch len(kv) {
+		case 1:
+			v, present := os.LookupEnv(kv[0])
+			if present {
+				vm.ExtVar(kv[0], v)
+			} else {
+				return nil, fmt.Errorf("Missing environment variable: %s", kv[0])
```
anguslees: Should we assume "" here instead of error?  What does `jsonnet` do in this case?

cmd/root.go
```diff
@@ -90,10 +97,68 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
...
+	}
+
+	extvarfiles, err := flags.GetStringSlice(flagExtVarFile)
+	if err != nil {
+		return nil, err
+	}
+	for _, extvar := range extvarfiles {
+		kv := strings.SplitN(extvar, "=", 2)
+		if len(kv) != 2 {
+			return nil, fmt.Errorf("Failed to parse ext-str-file: missing '=' in %s", extvar)
```
anguslees: Would be good to reuse flagExtVarFile constant in this error text, so I don't forget to update it if the constant ever changes..

cmd/root.go
```diff
@@ -90,10 +97,68 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
...
+
+	tlavarfiles, err := flags.GetStringSlice(flagTlaVarFile)
+	if err != nil {
+		return nil, err
+	}
+	for _, tlavar := range tlavarfiles {
+		kv := strings.SplitN(tlavar, "=", 2)
 		if len(kv) != 2 {
-			return nil, fmt.Errorf("Failed to parse extvar: missing '=' in %s", extvar)
+			return nil, fmt.Errorf("Failed to parse tla-str-file: missing '=' in %s", tlavar)
```
anguslees: Ditto: reuse flagTlaVarFile const

cmd/show_test.go
```diff
@@ -46,6 +47,8 @@ func TestShow(t *testing.T) {
   "number": 42,
   "string": "bar",
   "notAVal": "aVal",
+  "notAnotherVal": "aVal2",
+	"filevar": "foo",
```
anguslees: Odd indenting?

cmd/show_test.go
```diff
@@ -57,11 +60,16 @@ func TestShow(t *testing.T) {
...
+		os.Setenv("anVar", "aVal2")
+		defer os.Unsetenv("anVar")
+
 		output := cmdOutput(t, []string{"show",
 			"-J", filepath.FromSlash("../testdata/lib"),
 			"-o", format,
 			filepath.FromSlash("../testdata/test.jsonnet"),
 			"-V", "aVar=aVal",
+			"-V", "anVar",
+			"--ext-str-file", "filevar=../testdata/extvar.file",
```
anguslees: Use `filepath.FromSlash` to build this arg value, so we can pretend it works on Windows.

testdata/extvar.file
```diff
@@ -0,0 +1 @@
+foo
```
anguslees: Can you put a trailing newline in here (and update the test value accordingly)?

(Turning off my editor's trailing newline behaviour is such hard work - first world problems)

cmd/root.go
GauntletWizard: $ jsonnet foo.jsonnet -V BAR
ERROR: Environment variable BAR was undefined.

cmd/show_test.go
GauntletWizard: Yep. Looked right in my editor.

testdata/extvar.file
GauntletWizard: I guess my next PR will be to add a chomp function...

GauntletWizard:
Cool. What's the release schedule look like? I use your binaries in our
integration tests (path of least resistance); Looking forward to making use
of these.

anguslees:
v0.3.0 has been released with this (and your other) PRs :-)

Thanks again for the contributions.

PR #48, opened 2017-06-30
Publish lib/kubecfg.libsonnet in release

anguslees:
Makes sense for a user to have `kubecfg.libsonnet` easily available
without downloading the rest of the git repo.  Include in the release.


anguslees:
This feels wrong - it's equally easy just to point to the file via a tag-specific github URL.

I think the "real" fix is embedding the file in the executable, so we're back to a single download.   Abandoning this particular PR while I think about it more.

PR #53, opened 2017-07-06
Anonymously import the GCP auth provider to get the registration side effect.

tomwilkie:
Fixes #45 

Without this change, gcp auth was failing with `FATAL No Auth Provider found for name "gcp"`

Issue #46, opened 2017-06-29
Large diff output for missing default values


tomwilkie:
When I use `kubecfg diff` I tend to see a wall of red, representing default values missing from my local config but returned by the API server.  To work around this in weaveworks/kubediff I only check the local config is a subset of the config returned by the API server - ie we ignore fields which don't exist in the local config.

If this something we can do for kubecfg?

anguslees:
Yes, the diff output could do with a lot more cleanup (a real simple example would be removing the status fields, etc that we *know* aren't going to appear in local config).

I'm a bit nervous about naively applying your proposed approach: removing an entire sidecar container would (eg) be something I would want to clearly show up in a diff.  Probably have to put together some example situations and think about this a bit.

100% agreed on the need to remove spurious red noise.

tomwilkie:
With kubediff we found doing subset check on dictionaries, but a full equality check on lists works around that problem.  Its a bit of a hack.  Also, the API server is sometime "smart" about updates, and lists you read back from the API server might not be in the same order as the config you applied (for instance, I think this happens when adding a port to a service).  So we have special case handling for certain lists.

Does the swagger spec have default values in it?  If so we could use that to ignore fields that are missing but set to their defaults...

PR #54, opened 2017-07-06
Add 'subset' diff stategy option that checks the live state is a subset of the config state.

tomwilkie:
Part of #46, looking for feedback and discussion more than anything.

This PR adds a `--diff-strategy=subset` flag (default value is `all`, for the current behaviour).  The affect of this flag is to remove fields from the live state that are not present in the config state, reducing the diff to a subset operation.  This is how [weaveworks/kubediff](https://github.com/weaveworks/kubediff) behaves.  Note that for lists of differing lengths we preserve the live state, meaning you'll still get diff results for missing (or added) ports, containers etc.

IMO this approach turns out to be super useful - after all, if you haven't declared the desired value for a given field in you config, do you really care what its value is?  One limitation with this approach is that we won't catch labels or annotations that have been remove from config and not from live.

An alternative approach would be to "flesh out" the local config with default values for all the fields, remove the status value which will never be present locally, and then do a full diff.   This sounds super brittle to me, as you would need to keep you local kubecfg in sync with the cluster version (or make kubecfg cluster version aware).  We'd also need to maintain a black list of fields which are never going to be present locally but don't have sensible default values - for instance most of the stuff in metadata, such as `resourceVersion`, `selfLink`, `uid` etc.

WDYT?

(NB includes #53 until that is merged)


cmd/diff.go
```diff
@@ -31,7 +31,10 @@ import (
 	"github.com/ksonnet/kubecfg/utils"
 )
 
+var diffStrategy string
+
 func init() {
+	diffCmd.Flags().StringVar(&diffStrategy, "diff-strategy", "all", "Diff strategy, all or subset.")
```
anguslees: See the pattern used for other commands: Drop the `diffStrategy` global and use `flags.GetString()` below.
ie:

```go
const flagDiffStategy = "diff-strategy"

func init() {
    diffCmd.PersistentFlags().String(flagDiffStrategy, "all", "Diff strategy, all or subset.")
    // ...
}

var diffCmd = &cobra.Command{
   // ...
   RunE: func(...) {
      flags := cmd.Flags()
      diffStrategy, err := flags.GetString(flagDiffStrategy)
      if err != nil {
         return err
      }

      // ...
}
```

cmd/diff.go
```diff
@@ -103,6 +110,41 @@ var diffCmd = &cobra.Command{
 	},
 }
 
+func removeFields(config, live interface{}) interface{} {
```
anguslees: Can you add some tests for these?  In particular, I had to stare at `removeListFields` a bit before I was happy it was the desired behaviour (I agree it is) - some simple edge test cases might serve as documentation for the next person...

cmd/diff.go
```diff
@@ -103,6 +110,41 @@ var diffCmd = &cobra.Command{
...
+		if !ok {
+			continue
+		}
+		result[k] = removeFields(v1, v2)
+	}
+	return result
+}
+
+func removeListFields(config, live []interface{}) []interface{} {
+	result := []interface{}{}
```
anguslees: We know the length of the array, may as well use that fact:
```go
result := make([]interface{}, 0, len(live))
```

tomwilkie:
Thanks for the feedback @anguslees!  Have added a bunch of tests and fixed the other comments.

Issue #52, opened 2017-07-06
osx build is extremely flaky (jsonnet_cgo)


anguslees:
travis/osx auto tests fail about 4/5 times with a `fatal: morestack on g0` stacktrace from jsonnet_cgo code.

anguslees:
Hopefully addressed by #57

PR #57, opened 2017-07-07
Update to newer xcode for travis osx builds

anguslees:
Frequent `make test` crashes on (default) xcode7.3 in jsonnet_cgo. Use
a different/newer build environment.


anguslees:
From a small sample of test builds just now, this *appears* to fix #52 
Will keep a watch on future builds and reopen if we still see crashes.

PR #64, opened 2017-07-25
Make kubecfg diff exit with non-zero code when a diff is found.

### LINKED ISSUE #60: diff exit code could reflect presence of a diff
i.e. 1 means diff found, 0 when no diff.

tomwilkie:
This makes it easier to integrate into scripts that might automatically apply changes, for instance.

Fixes #60

### DIFF
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -67,6 +67,7 @@ var diffCmd = &cobra.Command{
 
 		sort.Sort(utils.AlphabeticalOrder(objs))
 
+		differencesFound := false
 		for _, obj := range objs {
 			desc := fmt.Sprintf("%s/%s", obj.GetKind(), fqName(obj))
 			log.Debugf("Fetching ", desc)
@@ -85,9 +86,10 @@ var diffCmd = &cobra.Command{
 			}
 
 			fmt.Fprintln(out, "---")
-			fmt.Fprintf(out, "- live %s\n+ config %s", desc, desc)
+			fmt.Fprintf(out, "- live %s\n+ config %s\n", desc, desc)
 			if liveObj == nil {
 				fmt.Fprintf(out, "%s doesn't exist on server\n", desc)
+				differencesFound = true
 				continue
 			}
 
@@ -98,6 +100,7 @@ var diffCmd = &cobra.Command{
 			diff := gojsondiff.New().CompareObjects(liveObjObject, obj.Object)
 
 			if diff.Modified() {
+				differencesFound = true
 				fcfg := formatter.AsciiFormatterConfig{
 					Coloring: istty(out),
 				}
@@ -112,6 +115,9 @@ var diffCmd = &cobra.Command{
 			}
 		}
 
+		if differencesFound {
+			return fmt.Errorf("Differences found.")
+		}
 		return nil
 	},
 }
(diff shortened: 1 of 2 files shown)


cmd/diff.go
```diff
@@ -112,6 +115,9 @@ var diffCmd = &cobra.Command{
 			}
 		}
 
+		if differencesFound {
+			return fmt.Errorf("Differences found.")
```
anguslees: I think callers will want to be able to check for this specific error condition, so put this in an exported global:
```go
var ErrDifferencesFound = errors.New("Differences found")
```
.. then I'm tempted to modify `main()` to look for this specific error and return a known exit code so calling scripts don't have to go grepping for a string.  Unfortunately diff(1) says "1 if different, 2 if trouble" and it's going to be hard to stop returning 1 for random fatal errors - perhaps we could invert that and return 2 in the "if different" case?

tomwilkie: +1 I like.

Turns out not to difficult to make "1 if different, 2 if trouble", as long as you're happy with 2 happening for trouble on other commands to?

tomwilkie:
CI is failing on Mac OS, but not in my code.  Is this a known issue?

cmd/diff.go
anguslees: My concern was some random code somewhere else that calls `log.Fatal()` or `panic()` directly.  I think it is more important to ensure that they fall into the "if trouble" exit code (whichever code that is) and less important that we match the exact diff(1) behaviour.

Now the bad news.  From a quick test just now:
- `log.Fatal` (logrus and standard go log versions) results in `exit(1)`
- `panic()` results in `exit(2)`

So that means we need to make "if different" something other than 1 or 2.  Perhaps 10 or something else clearly "special" (I don't see anything appropriate in sysexits.h).

anguslees:
> CI is failing on Mac OS, but not in my code. Is this a known issue?

cgo seems to be surprisingly flaky (sensitive to specific toolchain versions) on MacOS.  This specific example is new to me, and almost certainly unrelated to your change.  :(

main.go
```diff
@@ -32,7 +34,13 @@ func main() {
 		// errors, like invalid command line flags.
 		logFmt := cmd.NewLogFormatter(log.StandardLogger().Out)
 		log.SetFormatter(logFmt)
+		log.Error(err.Error())
 
-		log.Fatal(err.Error())
+		switch err {
```
anguslees: (Just propagating my comment against the older diff to here:)

My concern was some random code somewhere else that calls log.Fatal() or panic() directly. I think it is more important to ensure that they fall into the "if trouble" exit code (whichever code that is) and less important that we match the exact diff(1) behaviour.

Now the bad news. From a quick test just now:

log.Fatal (logrus and standard go log versions) results in exit(1)
panic() results in exit(2)
So that means we need to make "if different" something other than 1 or 2. Perhaps 10 or something else clearly "special" (I don't see anything appropriate in sysexits.h).

(Edit: .. or we could try to recover from panic() and turn it into an exit(1). That seems unnecessarily complex just to reclaim a specific exit value)

cmd/diff.go
tomwilkie: Ah okay I get you.  Will update.

PR #67, opened 2017-07-27
Add manifestJson and manifestYaml functions

### LINKED ISSUE #42: Want to build config maps that embed YAML.
I would love to be able to use jsonnet and kubecfg to build config maps that themselves contain YAML, which I would like to be native jsonnet object thats I can manipulate.  The concrete examples is embedding Prometheus config in a config map, and manipulating that config based on the environment.

To this end, I wrote a function that serialises an object to YAML (see https://github.com/tomwilkie/kubecfg/tree/unparse-yaml), but I hit the restriction "Native extensions can only take primitives." (https://github.com/google/jsonnet/blob/master/core/vm.cpp#L2157).

Is this something worth pursuing?  If so, I guess I'd need to make a PR into jsonnet to add a `std.manifestYaml(v)` like the existing manifest functions?

anguslees:
jsonnet's `std.toString()` returns "compact" JSON.  Sometimes this is
not appropriate, particularly when generating large output that might
need to be viewed/debugged by a human.

This change implements manifestYaml and manifestJson functions, which
generate YAML and "pretty" JSON respectively.

A slight implementation wrinkle is that libjsonnet currently only
supports passing primitive (scalar) types to native functions, so
these arguments need to be JSON-serialised/unserialised across the
native function call boundary.

Fixes #42

### DIFF
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -13,6 +13,12 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 
+// NB: libjsonnet native functions can only pass primitive types, so
+// some functions json-encode the arg.  These "*FromJson" functions
+// will be replaced by regular native version when libjsonnet is able
+// to support this.  This file strives to hide this implementation
+// detail.
+
 {
   // parseJson(data): parses the `data` string as a json document, and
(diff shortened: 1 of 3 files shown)


tomwilkie:
\o/

PR #72, opened 2017-08-02
Implement garbage collection

### LINKED ISSUE #15: Implement garbage collection of objects that no longer appear in config
Desired workflow:
- create some jsonnet config
- `kubecfg update` (should create new objects)
- remove some of the objects from jsonnet config
- `kubecfg update` (should remove "stale" objects)

Suggested implementation:
- Add some sort of "garbage-collect-me: true" annotation to all objects we create unless it already exists (perhaps with a false value)
- After regular update:
  - list all objects in all namespaces (!! this requires walking known object types, and will probably take a while)
  - delete all objects that have garbage-collect-me=true and were *not* in the update we just did

One possible optimisation is to only look in Namespaces that were mentioned in the update, but this is insufficient in the general case (will leak namespaces that are removed from config).

Must have a flag to disable it.  Current proposal is to _enable_ it by default.

anguslees:
Fixes #15

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -31,8 +31,10 @@ import (
 	"github.com/spf13/cobra"
 	jsonnet "github.com/strickyak/jsonnet_cgo"
 	"golang.org/x/crypto/ssh/terminal"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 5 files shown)


anguslees:
(cmd/update.go is now too big, will refactor in another PR)

PR #73, opened 2017-08-03
Report client-go version in `version` subcommand

anguslees:
Adds output like:
```
client-go version: v1.6.8-beta.0+$Format:%h$
```

(Apparently `$Format:%h$` is a misplaced feature - and requires us to declare the client-go git sha at link time, which is too awkward for us at this point.  The preceding version text is useful, however.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,34 @@
 language: go
 
 go:
-  - 1.7.x
   - 1.8.x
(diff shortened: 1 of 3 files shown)


anguslees:
Fwiw, this was broken by kubernetes/kubernetes#55353 :cry:

PR #75, opened 2017-08-08
Move code around and other cleanups

anguslees:
A number of long-overdue minor refactors - there should be very few observable changes:

- Ignore "appengine" build tag when vendoring
- Move all the non-command-line bulk of update into a library package
- Improve log description of object
- Move clientForResource into utils


anguslees:
(note to reviewers: probably want to review the commits individually, rather than as a merged stack)

hausdorff:
I'll take a look tomorrow.

utils/meta.go
```diff
@@ -71,3 +74,31 @@ func SetMetaDataAnnotation(obj metav1.Object, key, value string) {
...
 }
+
+// ResourceNameFor returns a lowercase plural form of a type, for
+// human messages.  Returns lowercased kind if discovery lookup fails.
+func ResourceNameFor(disco discovery.ServerResourcesInterface, o runtime.Object) string {
+	gvk := o.GetObjectKind().GroupVersionKind()
+	rls, err := disco.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
+	if err != nil {
+		log.Debugf("Discovery failed for %s: %s, falling back to kind", gvk, err)
+		return strings.ToLower(gvk.Kind)
```
hausdorff: I'm definitely not opposed to it, but I'm curious why we even bother?

cmd/update.go
```diff
@@ -71,33 +56,35 @@ var updateCmd = &cobra.Command{
...
 		}
 
-		dryRun, err := flags.GetBool(flagDryRun)
+		c.DryRun, err = flags.GetBool(flagDryRun)
 		if err != nil {
 			return err
 		}
 
-		clientpool, disco, err := restClientPool(cmd)
+		c.ClientPool, c.Discovery, err = restClientPool(cmd)
```
hausdorff: Not necessarily important for this PR, but:

I wonder if we should be moving stuff like `restClientPool` into another package as well. I think it would be nice to be able to pull the code that drives the application metadata management (which involves editing data on disk) into its own package, and in order to do schema caching, it seems like it would be useful to pull stuff like this into a package they mutually depend on, since it's used both here and in the `validate` patch.

utils/client.go
```diff
@@ -102,3 +107,47 @@ func (c *memcachedDiscoveryClient) SwaggerSchema(version schema.GroupVersion) (*
...
+	}
+
+	for _, r := range resources.APIResources {
+		if r.Kind == gvk.Kind {
+			log.Debugf("Chose API '%s' for %s", r.Name, gvk)
+			return &r, nil
+		}
+	}
+
+	return nil, fmt.Errorf("Server is unable to handle %s", gvk)
```
anguslees: Good suggestion.  I'll treat per-object errors as non-fatal during delete.  (Alternatively, I could add a `--force` flag that enabled this, and leave it as fatal by default)

utils/meta.go
anguslees: It's common to have different resources with the same names - eg: a service, deployment and configmap all called "nginx".  So I need to mention the kind somewhere in error messages / progress output, or else you just see "nginx" several times, with no idea which object that actually represents.  Originally I used just gvk.Kind, but I felt the resource name was a nicer form for humans (hyphenated, lowercase, pluralised).

cmd/update.go
anguslees: `restClientPool` is used in just about every subcommand, yes.  I moved it into utils/ so it could be used by other (non-command) golang code.  Happy to move it again if you would prefer a better-named package ("utils" is awfully nondescript).

utils/client.go
anguslees: (-> followup PR.  Since this PR is mostly about moving code around, I don't want to confuse it by introducing a new flag at the same time)

PR #76, opened 2017-08-08
Add application metadata management facilities

hausdorff:
The goal of this PR is to build assent about the general direction of the core application metadata management abstractions, and particularly to get early feedback so that we don't waste too much time implementing things we don't agree on at their core.

A ksonnet application will eventually depend heavily on a concrete directory structure, which is intended to help organize not only code but metadata about the application.

This PR will begin the process of abstracting over the state machine that manipulates this metadata (_i.e._, non-core-application constructs the user expects to never hand alter by themselves). For example, a user might wish to `get` a library or `search` for a prototype. In principle, the abstractions proposed here are meant to simplify the process of executing these commands as much as possible, in a way that is testable.

At this stage, the abstractions are fairly immature, but it should give a clear idea of the direction we're headed:

* [`metadata/interface.go`][1] exposes the interface `metadata.Manager` which is the core interface abstracting over application metadata management described above. The interface is empty now, but with possible signatures commented in to give an idea of where it's going.
* The concrete implementation of this interface, [`metadata/manager.go`][2] is fairly bare-bones at this point, essentially only setting up the directory structure. When the k8s changes are separated out of the `cmd/` module, we will be able to implement getting the OpenAPI spec from a server, but for now we only support getting such information from files.
* The implementation sits on top of Afero, a mockable FS. This is convenient mainly for testing.

[1]: https://github.com/ksonnet/kubecfg/blob/36de6056c844b45b2cb0e34529d5891a182da24e/metadata/interface.go
[2]: https://github.com/ksonnet/kubecfg/blob/36de6056c844b45b2cb0e34529d5891a182da24e/metadata/manager.go

### DIFF
--- a/metadata/clusterspec.go
+++ b/metadata/clusterspec.go
@@ -1,0 +1,43 @@
+package metadata
+
+import (
+	"github.com/spf13/afero"
+)
+
+type clusterSpecFile struct {
+	specPath AbsPath
+}
+
+func (cs *clusterSpecFile) data() ([]byte, error) {
+	return afero.ReadFile(appFS, string(cs.specPath))
+}
+
+func (cs *clusterSpecFile) resource() string {
+	return string(cs.specPath)
+}
+
+type clusterSpecLive struct {
+	apiServerURL string
+}
+
+func (cs *clusterSpecLive) data() ([]byte, error) {
+	// TODO: Implement getting spec from path, k8sVersion, and URL.
+	panic("Not implemented")
+}
+
+func (cs *clusterSpecLive) resource() string {
+	return string(cs.apiServerURL)
+}
+
+type clusterSpecVersion struct {
+	k8sVersion string
+}
+
+func (cs *clusterSpecVersion) data() ([]byte, error) {
+	// TODO: Implement getting spec from path, k8sVersion, and URL.
+	panic("Not implemented")
+}
+
+func (cs *clusterSpecVersion) resource() string {
+	return string(cs.k8sVersion)
+}
--- a/metadata/interface.go
+++ b/metadata/interface.go
@@ -1,0 +1,80 @@
+package metadata
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/afero"
+)
+
+// AbsPath is an advisory type that represents an absolute path. It is advisory
+// in that it is not forced to be absolute, but rather, meant to indicate
+// intent, and make code easier to read.
+type AbsPath string
+
+// Manager abstracts over a ksonnet application's metadata, allowing users to do
+// things like: create and delete environments; search for prototypes; vendor
+// libraries; and other non-core-application tasks.
+type Manager interface {
+	Root() AbsPath
+	//
+	// TODO: Fill in methods as we need them.
+	//
+	// GetPrototype(id string) Protoype
+	// SearchPrototypes(query string) []Protoype
+	// VendorLibrary(uri, version string) error
+	// CreateEnv(name string, spec *ClusterSpec) error
+	// DeleteEnv(name string) error
+	//
+}
+
+// Find will recursively search the current directory and its parents for a
+// `.ksonnet` folder, which marks the application root. Returns error if there
+// is no application root.
+func Find(path AbsPath) (Manager, error) {
+	return findManager(path, afero.NewOsFs())
+}
+
+// Init will retrieve a cluster API specification, generate a
+// capabilities-compliant version of ksonnet-lib, and then generate the
+// directory tree for an application.
+func Init(rootPath AbsPath, spec ClusterSpec) (Manager, error) {
+	return initManager(rootPath, spec, afero.NewOsFs())
+}
+
+// ClusterSpec represents the API supported by some cluster. There are several
+// ways to specify a cluster, including: querying the API server, reading an
+// OpenAPI spec in some file, or consulting the OpenAPI spec released in a
+// specific version of Kubernetes.
+type ClusterSpec interface {
+	data() ([]byte, error)
+	resource() string // For testing parsing logic.
+}
+
+// ParseClusterSpec will parse a cluster spec flag and output a well-formed
+// ClusterSpec object. For example, if the flag is `--version:v1.7.1`, then we
+// will output a ClusterSpec representing the cluster specification associated
+// with the `v1.7.1` build of Kubernetes.
+func ParseClusterSpec(specFlag string) (ClusterSpec, error) {
+	split := strings.SplitN(specFlag, ":", 2)
+	if len(split) == 0 || len(split) == 1 || split[1] == "" {
+		return nil, fmt.Errorf("Invalid API specification '%s'", specFlag)
+	}
+
+	switch split[0] {
+	case "version":
+		return &clusterSpecVersion{k8sVersion: split[1]}, nil
+	case "file":
+		abs, err := filepath.Abs(split[1])
+		if err != nil {
+			return nil, err
+		}
+		absPath := AbsPath(abs)
+		return &clusterSpecFile{specPath: absPath}, nil
+	case "url":
+		return &clusterSpecLive{apiServerURL: split[1]}, nil
+	default:
+		return nil, fmt.Errorf("Could not parse cluster spec '%s'", specFlag)
+	}
+}
--- a/metadata/manager.go
+++ b/metadata/manager.go
@@ -1,0 +1,138 @@
+package metadata
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+func appendToAbsPath(originalPath AbsPath, toAppend ...string) AbsPath {
+	paths := append([]string{string(originalPath)}, toAppend...)
+	return AbsPath(path.Join(paths...))
+}
+
+const (
+	defaultEnvName = "dev"
+
+	ksonnetDir    = ".ksonnet"
+	libDir        = "lib"
+	componentsDir = "components"
+	vendorDir     = "vendor"
+	schemaDir     = "vendor/schema"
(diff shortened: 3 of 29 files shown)


metadata/clusterspec.go
```diff
@@ -0,0 +1,87 @@
...
+	k8sVersion *string
+	specPath   *AbsPath
+	specURL    *string
+	specData   []byte
+}
+
+func ParseClusterSpec(specFlag string) (*ClusterSpec, error) {
+	separator := ":"
+	split := strings.Split(specFlag, separator)
+	if len(split) == 0 || len(split) == 1 || split[1] == "" {
```
hausdorff: Technically this conditional is wrong since something like `"file::foo"` is ok on some filesystems. I'll fix it later.

metadata/clusterspec.go
```diff
@@ -0,0 +1,87 @@
...
+		// TODO: Implement getting spec from path, k8sVersion, and URL.
+		panic("Not implemented")
+	}
+}
+
+// TODO: Put something like this in a k8s package when Gus merges his
+// "modularization" PR.
+
+// func downloadSchema(groupVersion string) ([]byte, error) {
+// 	prefix := "apis"
```
anguslees: This function can be replaced with `disco.SwaggerSchema(gv)`

metadata/clusterspec.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func ParseClusterSpec(specFlag string) (*ClusterSpec, error) {
+	separator := ":"
+	split := strings.Split(specFlag, separator)
+	if len(split) == 0 || len(split) == 1 || split[1] == "" {
+		return nil, fmt.Errorf("Invalid API specification '%s'", specFlag)
+	}
+
+	switch split[0] {
```
anguslees: If `ClusterSpec` is supposed to be a sum type (ie: only one of these is valid at any point in time), I suspect you instead want 3 separate golang types that implement a common interface.

metadata/clusterspec.go
```diff
@@ -0,0 +1,87 @@
...
+type ClusterSpec struct {
+	k8sVersion *string
+	specPath   *AbsPath
+	specURL    *string
+	specData   []byte
+}
+
+func ParseClusterSpec(specFlag string) (*ClusterSpec, error) {
+	separator := ":"
+	split := strings.Split(specFlag, separator)
```
anguslees: Use `strings.SplitN(specFlag, ":", 2)` here, and remove all the `strings.Join(split[1:], ":")` calls below.

metadata/clusterspec.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(split) == 0 || len(split) == 1 || split[1] == "" {
+		return nil, fmt.Errorf("Invalid API specification '%s'", specFlag)
+	}
+
+	switch split[0] {
+	case "version":
+		return newK8sVersionSpec(strings.Join(split[1:], separator)), nil
+	case "file":
```
anguslees: I'm not 100% sure where this is going, but I suggest dropping the "version" case (there's no benefit to having pre-canned configs vs just querying a live server imo), and just rolling the "file" case into the "url" case by making it always a URL that is relative to "file://$pwd".  Iiuc, that leaves us with *only* the "url" case, so we can remove most of the following code and simplify the user-visible "cluster spec" format to just a URL.

metadata/manager.go
```diff
@@ -0,0 +1,164 @@
...
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+func appendToAbsPath(originalPath AbsPath, toAppend ...string) AbsPath {
+	paths := append([]string{string(originalPath)}, toAppend...)
+	return AbsPath(path.Join(paths...))
```
anguslees: No need to build an intermediate array, just:
```go
return AbsPath(path.Join(originalPath, toAppend...))
```

metadata/clusterspec.go
hausdorff: It's supposed to be a Haskell-style sum type, yes. I felt like, as long as it's not exposed to users except with these specific constructors, but I'm like a 3/10 on this, so let's do it your way.

metadata/clusterspec.go
hausdorff: Do you feel strongly about making `file:` and `url:` the same? I'm probably a 7/10 in thinking that having the `file:` is nice sugar -- and besides, RFC 1738's spec for how file URLs work is not trivial to understand (_e.g._, `file:///thing` and `file://thing` actually mean pretty different things), which makes me hesitant to insist on using them.

hausdorff: Also I was thinking the `version:` tag would retrieve the schema from tagged releases in the k8s official repository.

metadata/manager.go
hausdorff: I think it would have to be something like:

```go
return AbsPath(path.Join(append([]string{string(originalPath)}, toAppend...)...))
```

I'm like a 2/10 on this, so I'm happy to do it this way if you really think that's cleaner, but I personally think it's worse.

metadata/interface.go
```diff
@@ -0,0 +1,97 @@
...
+// Init will retrieve a cluster API specification, generate a
+// capabilities-compliant version of ksonnet-lib, and then generate the
+// directory tree for an application.
+func Init(rootPath AbsPath, spec ClusterSpec) (Manager, error) {
+	return initManager(rootPath, spec)
+}
+
+// ClusterSpecType represents the set of ways to specify a cluster (and,
+// therefore, from what schema we should generate ksonnet-lib).
+type ClusterSpecType int
```
anguslees: This (and related references below) should not be necessary anymore.  If the caller really wants to know which implementation of the `ClusterSpec` interface they're dealing with then they can use a golang type-switch (and it probably also implies that the `ClusterSpec` interface is insufficient somewhere).

metadata/manager.go
```diff
@@ -0,0 +1,163 @@
...
+
+		lastBase = currBase
+		currBase = filepath.Dir(currBase)
+		if lastBase == currBase {
+			return nil, fmt.Errorf("No ksonnet application found")
+		}
+	}
+}
+
+func initManager(rootPath AbsPath, spec ClusterSpec) (*manager, error) {
```
anguslees: This only needs the spec data and not the full ClusterSpec object, so it will be easier to test, etc if it just accepts `data []byte` directly.

anguslees: oh, unless there's meant to be more here in future?  I'm unsure which of these functions are intended to be public vs private - since they're all lowercase at present.

metadata/manager.go
```diff
@@ -0,0 +1,163 @@
...
+	libDir        = "lib"
+	componentsDir = "components"
+	vendorDir     = "vendor"
+	schemaDir     = "vendor/schema"
+	vendorLibDir  = "vendor/lib"
+
+	schemaFilename = "swagger.json"
+)
+
+type manager struct {
```
anguslees: `appFS` should be a property on the manager rather than a global (so alternate implementations can be provided for testing, etc)

metadata/manager.go
```diff
@@ -0,0 +1,163 @@
...
+		schemaDir:      appendToAbsPath(rootPath, schemaDir),
+		vendorLibDir:   appendToAbsPath(rootPath, vendorLibDir),
+	}
+}
+
+func (m *manager) Root() AbsPath {
+	return m.rootPath
+}
+
+func (m *manager) cacheClusterSpec(name string, spec ClusterSpec) error {
```
anguslees: I suggest removing this function, unless there's an actual need for it somewhere.  The function signature implies that it caches the `ClusterSpec`, whereas it just takes a snapshot of the current spec data - in particular an external reference to URL/file won't be re-fetched in the future if the external resource changes.

metadata/manager.go
```diff
@@ -0,0 +1,163 @@
...
+	var err error
+
+	exists, err := afero.DirExists(appFS, string(m.rootPath))
+	if err != nil {
+		return fmt.Errorf("Could not check existance of directory '%s':\n%v", m.rootPath, err)
+	} else if exists {
+		return fmt.Errorf("Could not create app; directory '%s' already exists", m.rootPath)
+	}
+
+	err = appFS.MkdirAll(string(m.rootPath), os.ModePerm)
```
anguslees: It's a micro-cleanup, but once this merges we're ~never going to come back to this code:

```go
   paths := []AbsPath{
      m.rootPath,
      m.ksonnetPath,
      m.libPath,
      m.componentsPath,
      m.vendorDir,
      m.schemaDir,
      m.vendorLibDir,
   }
   for _, p := range paths {
      if err = appFS.MkdirAll(string(p), os.ModePerm); err != nil {
         return err
      }
   }

```

metadata/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	appPath := AbsPath("/fromEmptySwagger")
+	_, err = initManager(appPath, spec)
+	if err != nil {
+		t.Fatalf("Failed to init cluster spec: %v", err)
+	}
+
+	exists, err := afero.DirExists(appFS, string(appendToAbsPath(appPath, ksonnetDir)))
+	if !exists || err != nil {
+		t.Fatalf("Expected to create root app directory '%s', but directory not found", appPath)
```
anguslees: include `err` in the error message (here and below) rather than guessing at the reason.  (if the reported error is `nil`, then we know it was `!exists`)

metadata/manager.go
hausdorff: `Init` is meant to be public, `initManager` is meant to be a private testable implementation of `Init`.

I actually find this to be more clear as a typed `ClusterSpec` though, I kind of find naked `[]byte` to be an anti-pattern in the same sense as `void*`. I also don't see any serious degradation of testability, so if you don't mind I'd like to keep it.

PR #77, opened 2017-08-08
Implement `validate` subcommand

### LINKED ISSUE #18: Implement `check` subcommand
`kubecfg check` should perform as many validations as we can without sending the objects to the API server.

Specifically:
- evaluate the jsonnet (verifying syntax and jsonnet asserts)
- compare resulting json against swagger spec

Note: Unlike kubectl, swagger spec should (optionally?) be able to be read from static files on disk (ie: with *no* access to API server at runtime)

anguslees:
Fetches schema from server without caching.  Caching / offline
operation will be added in a followup PR.

Fixes #18

### DIFF
--- a/utils/openapi.go
+++ b/utils/openapi.go
@@ -1,0 +1,336 @@
+// Copyright 2017 The kubecfg authors
+//
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+
+	"github.com/emicklei/go-restful/swagger"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


utils/openapi.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/emicklei/go-restful/swagger"
+	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+)
+
+// This is unashamedly copied from
+// k8s.io/kubernetes/pkg/api/validation/schema.go, and rebased onto
+// client-go/apimachinery object types
```
hausdorff: The implications of maintaining this are unclear to me. Is the only reason we're copy+paste'ing this to deal with CRDs/TPRs? How do we expect this code to evolve over time?

anguslees: The conceptual contract here is that the k8s server is publishing an openapi-conformant spec, and we're using that to validate some local golang/json objects.  So the overall goal of this code is very unlikely to change - only as upstream k8s evolves their public openapi-spec interface.  The specific code might need to change in a minor way as we (or upstream) discover bugs, etc.

The reason it is copy+pasted is that the original code comes from *k8s* itself, and not from one of the published client libraries (client-go, apimachinery, etc).  Consequently, the original code depends on a whole lot of other k8s-internal types and packages that we don't need and can't use.  So we need to modify the original anyway.  I _could_ write an effectively brand new bit of code to walk the swagger API "inspired" by the upstream example, but intentionally keeping the kubecfg version close-ish to the literal k8s version felt like the better maintenance gamble...

.. And yes, the reason we're doing any of this vs using garethr's library is to deal with CRDs/TPRs and other dynamic k8s server behaviour like non-default alpha APIs, odd local code patches, etc.

PR #88, opened 2017-08-23
Implement `init`; bring synchronization primitives to spec

hausdorff:


hausdorff:
Marking this as [WIP] as I push up some more speculative stuff for Jessica Yuen.

metadata/clusterspec.go
```diff
@@ -1,15 +1,49 @@
...
 	"github.com/spf13/afero"
 )
 
+const (
+	k8sVersionURLTemplate = "https://raw.githubusercontent.com/kubernetes/kubernetes/%s/api/openapi-spec/swagger.json"
+)
+
+func parseClusterSpec(specFlag string, fs afero.Fs) (ClusterSpec, error) {
+	split := strings.SplitN(specFlag, ":", 2)
+	if len(split) == 0 || len(split) == 1 || split[1] == "" {
```
jessicayuen: if len(split) <= 1 || split[1] == "" { ?

cmd/init.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+var initCmd = &cobra.Command{
+	Use:   "init <app-name>",
+	Short: "Initialize a ksonnet project",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		flags := cmd.Flags()
+		if len(args) != 1 {
+			return fmt.Errorf("'init' requires a name for the application we're initializing")
+		}
```
jessicayuen: message might be misleading for `init My Project`; do you want to handle these cases now?

hausdorff: Resolved.

metadata/clusterspec.go
hausdorff: Resolved.

cmd/root.go
```diff
@@ -155,176 +152,49 @@ func (f *logFormatter) Format(e *log.Entry) ([]byte, error) {
...
-}
 
-func buildResolver(cmd *cobra.Command) (utils.Resolver, error) {
-	flags := cmd.Flags()
-	resolver, err := flags.GetString(flagResolver)
-	if err != nil {
-		return nil, err
-	}
-	failAction, err := flags.GetString(flagResolvFail)
+	spec.Resolver, err = flags.GetString(flagResolver)
```
anguslees: ? The resolver stuff has nothing to do with jsonnet (it's for docker tag->digest conversion).  Why are you moving it to the Expander abstraction ?

template/expander.go
```diff
@@ -0,0 +1,122 @@
...
+	ExtVars     []string
+	ExtVarFiles []string
+	TlaVars     []string
+	TlaVarFiles []string
+
+	Resolver   string
+	FailAction string
+}
+
+func (spec *Expander) Expand(paths []string) ([]*unstructured.Unstructured, error) {
```
anguslees: This is an awkward library API.  This function now builds it's own file paths, does file accesses, looks up environment variables, etc, which makes it very hard to reuse this code in any other situation (even unittests).  Can we define a narrower scope for what an "Expander" actually does, and refactor some of these environmental dependencies back out of this package?

template/resolver.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/ksonnet/kubecfg/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+func (spec *Expander) buildResolver() (utils.Resolver, error) {
```
anguslees: repeating earlier comment, this doesn't belong here - these are implementation details of the resolveImage() native jsonnet function and should be moved closer to the other native functions.

PR #93, opened 2017-08-30
Refactor cmd/ application logic to pkg/

jessicayuen:
This commit will separate the application level logic for the 'show', 'delete',
'validate', and 'diff' commands from the Cobra logic in the cmd/ package.
The application level logic will be placed in pkg/kubecfg/.

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -16,15 +16,9 @@
 package cmd
 
 import (
-	"fmt"
-	"sort"
-
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/ksonnet/kubecfg/utils"
+	"github.com/ksonnet/kubecfg/pkg/kubecfg"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


cmd/delete.go
```diff
@@ -42,75 +39,40 @@ var deleteCmd = &cobra.Command{
...
-			err = c.Delete(obj.GetName(), &deleteOpts)
-			if err != nil && !errors.IsNotFound(err) {
-				return fmt.Errorf("Error deleting %s: %s", desc, err)
-			}
-
-			log.Debugf("Deleted object: ", obj)
-		}
-
-		return nil
+		return c.Run(metadata.AbsPath(cwd))
```
anguslees: I'm with you right up to here - why is $cwd a required argument for delete? (haven't we already done all the context-specific file operations already in preceding lines?)
(and similar for other files)

pkg/kubecfg/delete.go
```diff
@@ -0,0 +1,90 @@
...
+	Discovery        discovery.DiscoveryInterface
+	DefaultNamespace string
+
+	Expander *template.Expander
+	Files    []string
+
+	GracePeriod int64
+}
+
+func (c DeleteCmd) Run(wd metadata.AbsPath) error {
```
anguslees: .. and indeed, `wd` is unused?

pkg/kubecfg/delete.go
```diff
@@ -0,0 +1,90 @@
...
+	version, err := utils.FetchVersion(c.Discovery)
+	if err != nil {
+		return err
+	}
+
+	objs, err := c.Expander.Expand(c.Files)
```
anguslees: Many commands have these same basic two stages:
1. acquire a bunch of API objects, from reading various files, evaluating jsonnet, etc
2. do something with those API objects

Here, you've moved (1) into the function, rather than leaving it as a separate refactored stage (I had it in a generic `readObjs` function iirc).
Are we better leaving (1) outside these functions?  (ie: move all the Expander+files logic to some earlier function, and just pass API objects to `DeleteCmd`)

cmd/delete.go
jessicayuen: thanks for the catch -- oversight on my part when a took some of the logic from `update`

pkg/kubecfg/delete.go
jessicayuen: I wasn't able to find a generic readObjs function (please correct me if I'm wrong), but I will write one and extract (1) from the commands (including `update` which has similar logic in `master`) in a separate commit. Will also pass the API Objects instead of Expander & Files to each Cmd type as you mentioned.

hausdorff: Wrapping up this discussion and the slack discussion for posterity, we've agreed to:

1. Keep the `template.Expander` abstraction, since it will be useful for things like servers.
2. But run it from the `cmd/` package, passing the API objects to the `pkg/`. Possibly using the old function names, `readObjs`

cmd/root.go
```diff
@@ -269,3 +269,17 @@ func getFiles(cmd *cobra.Command, args []string) ([]string, error) {
 
 	return kubecfg.GetFiles(metadata.AbsPath(cwd), env, files)
 }
+
+func readObjs(cmd *cobra.Command, args []string) ([]*unstructured.Unstructured, error) {
```
hausdorff: `cmd#getFiles` seems to be called only in one place (_i.e._, here), so it seems like it might be worth just deleting it and moving the contents here.

hausdorff:
LGTM, I'd say we should squash the two commits together before merging though.

pkg/kubecfg/delete.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2017 The kubecfg authors
```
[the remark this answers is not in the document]
anguslees: It's mean to be on all of them (see "how to use" in our own LICENSE file).  But I forgot some, and don't have a unittest to remind me...

So: please add to anywhere you notice.

cmd/delete.go
```diff
@@ -42,75 +36,30 @@ var deleteCmd = &cobra.Command{
...
+		c := kubecfg.DeleteCmd{}
 
-		files, err := getFiles(cmd, args)
+		c.GracePeriod, err = flags.GetInt64(flagGracePeriod)
 		if err != nil {
 			return err
 		}
 
-		vm, err := newExpander(cmd)
+		c.APIObjects, err = readObjs(cmd, args)
```
anguslees: I have a weak stylistic preference for making the objects be actual golang arguments to `Run()` - but that's because I think of the rest as setting up the "environment", and then potentially you could reuse the same `DeleteCmd` instance with different sets of objects to `Run(objs)`.

I'm also ok with not doing that, and the assumption is that `Run()` only gets called once - and all the rest is just a fancy way of passing many arguments to that single function.

hausdorff: I also prefer this. But I'm a 2/10 in terms of strength of this belief.

jessicayuen: Done.

PR #94, opened 2017-08-30
Fix README

mmikulicic:
Now the commands take either an env or a file, so let's at least fix the
README to make it work with files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,21 +43,21 @@ avoid an immediate `Killed: 9`.
 % export KUBECFG_JPATH=/path/to/kubecfg/lib
 
 # Show generated YAML
-% kubecfg show -o yaml examples/guestbook.jsonnet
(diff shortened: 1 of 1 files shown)


anguslees:
This is one of the annoying things about github - the frontpage README.md is from `master`, and not the most recent release.  So I have to break one or the other :(

mmikulicic:
yeah; there is an option to deal with that:

1. create a `release` branch, make it the default branch.
2. when making a new release merge master in the release branch and then tag.

Pros:
a. github will show the released readme; it's quite easy to switch the github UI to another branch
b. go get (for go >=1.6) will fetch the default github branch, i.e. people won't be building master accidentally.

Cons:
a. people might send PRs against `release`; hopefully it's obvious from the name that that would be wrong.
b. one could forget to keep `release` in sync with the latest release; nothing that some tooling can't help; but it makes everything more complicated.

PR #98, opened 2017-08-31
Build using go-1.9.x

anguslees:
- Build with both 1.9.x and 1.8.x.
- No longer test builds with golang 1.7.x.
- Continue to ignore failures from osx/1.8.x + osx/1.9.x
- Build releases with 1.9.x
- Update README.md to reflect >= go-1.8 requirement

Required various $(GO_FLAGS) refactorings, because `go vet` from
go-1.9 gets upset by `-ldflags` (previously ignored).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 language: go
 
 go:
+  # NB: order matters - matrix items that don't specify will use the
+  # first value (ditto for `os` below)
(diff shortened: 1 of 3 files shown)


hausdorff:
@anguslees Do we want to rebase on `master` and try a rebuild? I'm happy to merge it once that's through.

PR #100, opened 2017-09-01
Introduce initial version of the `prototype` command

hausdorff:


jessicayuen:
Do you want to add the License on top of these files?

prototype/snippet/parser_test.go
```diff
@@ -0,0 +1,430 @@
...
+	m := p.parse(value, false, false)
+	assertMarker(t, *m, ctors...)
+}
+
+func assertTextAndMarker(t *testing.T, value, escaped string, ctors ...marker) {
+	assertText(t, value, escaped)
+	assertMarkerValue(t, value, ctors...)
+}
+
+// 	function assertEscaped(value: string, expected: string) {
```
jessicayuen: did you mean to include this change?

hausdorff:
As @jessicayuen mentioned, there was enough code here that it was easier to sit down and go through the review together. For posterity, I've aggregated the feedback from Jess and others, as well as how it was addressed:

* `prototype` should work outside of an application.
* `prototype` subcommands (_i.e._, `use`, `describe`, and `init`) should have their own help messages; right now only the `prototype` root command has a help message.
* It would be nice to have an `prototype list` command to list all available prototypes.
  * **Deferred to another PR:** outside the scope of this PR.
* Fix the tests to include both system prototypes and novel, added prototypes.
* `metadata.Manager` does not need a search command right now; we only use the hard-coded prototypes right now, so just use the `prototype.Index`. 
* Fix up the system prototypes, which are kind of janky.
* Add a `type` field to the prototype schema (_i.e._, `YAML|JSON|Jsonnet`), and display it during `prototype search`, so that users know what it is.
  * **Deferred to another PR:** outside the scope of this PR.

In addition, I had to change the dates on a bunch of the commits to get the correct order in GH, so it looks like they were all added in the last iteration. There's got to be a better way.

hausdorff:
The integration tests are failing, but they are failing on master as well. So I'm going to check merge this and then try to fix those.

PR #105, opened 2017-09-03
Deprecate `update` subcommand; introduce `apply`

hausdorff:
This commit is a follow-up to the discussion in the ksonnet.next design
doc, in which users consistently expressed their preference that the
`update` command be called `apply`.


jessicayuen:
Approved this change as most of the work here is renaming update to apply, LGTM.

Issue #96, opened 2017-08-31
`lib/` in the app directory should bind be added to `-J` paths automatically.


hausdorff:

hausdorff:
Resolved in #106.

PR #106, opened 2017-09-04
Include `lib/` and `vendor/lib` when expanding templates

hausdorff:
Currently, if the user wants to deploy a ksonnet application, and that
application uses some vendored library, or something in the `lib/`
directory, they will have to pass the appropriate `-J` flags into the
command themselves.

This commit will automatically add these whenever we're in an app
directory and a command is issued. In particular, even if the user
passes in the `-f` flag (rather than an environment name), we'll still
add library paths to the command if we're in a ksonnet app directory.
This is meant to capture the case that a user wants to update one
resource in particular in a ksonnet application.

This commit will resolve issue #96.


cmd/root.go
```diff
@@ -245,10 +244,14 @@ func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.Discovery
 	return pool, discoCache, nil
 }
 
+// addEnvCmdFlags adds the flags that are common to the family of commands
+// whose form is `[<env>|-f <file-name>]`, e.g., `update` and `delete`.
```
jessicayuen: Might want to change `update` to `apply` here to be consistent with #105

cmd/root.go
```diff
@@ -245,10 +244,14 @@ func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.Discovery
...
 func addEnvCmdFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringArrayP(flagFile, flagFileShort, nil, "Filename or directory that contains the configuration to apply (accepts YAML, JSON, and Jsonnet)")
 }
 
+// parseEnvCmd parses the family of commands that come in the form `[<env>|-f
+// <file-name>]`, e.g., `update` and `delete`.
```
jessicayuen: same with here and a few of the following comments

cmd/root.go
```diff
@@ -281,10 +289,40 @@ func readObjs(cmd *cobra.Command, args []string) ([]*unstructured.Unstructured,
...
+	// Get all filenames that contain templates to expand. Importantly, we need to
+	// enforce the form `[<env-name>|-f <file-name>]`; that is, we need to make
+	// sure that the user either passed an environment name or a `-f` flag.
+	//
+
+	envPresent := env != nil
+	filesPresent := len(fileNames) > 0
+
+	// This is equivalent to: `if !xor(envPresent, filesPresent) {`
+	if envPresent && filesPresent {
```
jessicayuen: We discussed about potentially removing the mutual exclusivity of file and env. Did you want that change reflected in this PR?

hausdorff: I think we should move this to the env changes you have incoming.

cmd/root.go
hausdorff: Good point, but I believe we should do it in #105 instead (since this review does not depend on that one). I'll merge this, rebase that one, and then merge that one when it's approved.

PR #109, opened 2017-09-05
Support import param rewrite for jsonnet snippets

jessicayuen:
This commit will rewrite all parameters in the passed Jsonnet file of the form
`import param://foo` to `${foo}`. This will be a utility for `ksonnet prototype
use` to support LSP-spec-compliant snippet JSONs.


anguslees:
Why hijack `import` like this rather than (eg) provide a new function-like builtin, or use `std.extVar()` or similar existing mechanisms?

hausdorff:
So, there was some discussion on the kubernetes#ksonnet channel about the `import` syntax, which I think Dave Cunningham sums up well [here](https://github.com/ksonnet/kubecfg/pull/106).

To re-summarize, I think we are less comfortable with replacing calls to something like `std.extVar` because in general it can't be decided whether an expression is equivalent to a call to `std.extVar`. Consider, for example, `std["ext" + "Var"]`, which is fairly benign from a static analysis perspective, but enough to illustrate that this might be fairly complex.

In contrast, the advantage to using `import` is that it is completely decidable, totally unambiguous, and semantically a lot like what we intend to happen. It _is_ little ugly, but I (we?) think that's ok, since we expect there will be many fewer prototype authors than prototype users.

Anyway, @anguslees told us he was ok with it too (though "ok with it" may just mean "he got tired of talking about it" :) ).

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
+// Copyright 2017 The kubecfg authors
```
hausdorff: What do you think about moving `jsonnet_snippet.go` and `jsonnet_snippet_test.go` into `pkg/prototype/jsonnetSnippet/snippet.go`, _etc_.?

Also in the rest of the project I believe we're inconsistent about the case we're using for files, so I think we should just adopt `snake_case` as you have here.

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Replace all parameter imports
+	return replace(jsonnet, imports), nil
+}
+
+// ---------------------------------------------------------------------------
+
+func visit(node ast.Node, imports *[]ast.Import) {
+	switch n := node.(type) {
+	case *ast.Import:
```
hausdorff: Hmm, I think we're missing:

* `ImportStr`
* `Dollar`
* `LiteralBoolean`
* `LiteralNull`
* `LiteralNumber`
* `LiteralString`
* `Self`
* `Var`

I think this is ok, (since none of them have members we need to visit, but I think we should also explicitly decide whether we want to error out if we find a node with an unrecognized type.

My suspicion is yes, since if we add syntax to Jsonnet (as we did recently for `SuperIn`), then it will be possible for a user to write `import "params://"` and have it not picked up, because we don't know how to visit the children.

So, I recommend that we make the above nodes explicitly handled by the `switch`, but do nothing when we visit them. Then, we might have a `default` that errors out if it doesn't recognize the node type.

Thoughts?

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// replace converts all parameters in the passed Jsonnet of form
+// `import 'param://port'` into `${port}`.
+func replace(jsonnet string, imports []ast.Import) string {
+	lines := strings.Split(jsonnet, "\n")
+
+	// Imports must be sorted by location so that replacement can start from
+	// the last import node.
+	sort.Sort(byLocation(imports))
```
hausdorff: Could use `sort.Slice` also. Then you don't have to define `Len`, `Swap`, and `Less`.

hausdorff: Also, might want to call out that we're sorting in reverse order, and why.

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
...
+	return len(a)
+}
+
+func (a byLocation) Swap(i, j int) {
+	a[i], a[j] = a[j], a[i]
+}
+
+func (a byLocation) Less(i, j int) bool {
+	if a[i].Loc().End.Line == a[j].Loc().End.Line {
+		return a[i].Loc().End.Column > a[j].Loc().End.Column
```
hausdorff: I'm reasonably sure this doesn't matter but I want to call it out to see if you can think of a case where this fails.

If a bunch of nodes end at the same line and column, it's not clear to me what the behavior here is. We need to be sure that we're convinced that it doesn't matter if several nodes ending at the same column on the same line get more or less randomly reordered.

Since `import` nodes can't be nested, I think it does not matter.

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
...
+
+const (
+	paramPrefix            = "param://"
+	paramReplacementPrefix = "${"
+	paramReplacementSuffix = "}"
+)
+
+// Parse rewrites the imports in a Jsonnet file before returning the parsed
+// TextMate snippet.
+func Parse(fn string, jsonnet string) (snippet.Template, error) {
```
hausdorff: I think we probably want to return a `*SpecificationSchema` here.

But that implies parsing the comments to populate those fields, so I'm fine booting that to another PR.

prototype/jsonnet_snippet_test.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func TestParse(t *testing.T) {
+	tests := []struct {
+		jsonnet  string
+		expected string
+	}{
+		// Test multiple import param replacement in a Jsonnet file.
+		{
+			`
```
hausdorff: Would be nice to see a couple more tests exercising things like "no parameters found" and stuff like that.

prototype/jsonnet_snippet.go
jessicayuen: Will move to `prototype/snippet/jsonnet/snippet.go`.

prototype/jsonnet_snippet.go
```diff
@@ -0,0 +1,222 @@
...
+//    limitations under the License.
+
+package prototype
+
+import (
+	"sort"
+	"strings"
+
+	ast "github.com/google/go-jsonnet/ast"
+	parser "github.com/google/go-jsonnet/parser"
```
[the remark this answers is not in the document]
jessicayuen: I did not know that, so thanks! Will fix.

prototype/jsonnet_snippet.go
jessicayuen: This makes sense to me, as we do probably want to error out for invalid Jsonnet input.

prototype/jsonnet_snippet.go
jessicayuen: The intention was that all the nodes being sorted would be `Import` type nodes, which I assumed could not have an overlap in Location or be nested. Were you thinking the sort should support other node types for future-proofing?

hausdorff: Oh, derp, you're right. Ignore.

prototype/jsonnet_snippet.go
jessicayuen: Both these are done.

prototype/jsonnet_snippet.go
jessicayuen: We're deciding to address this in a future PR.

prototype/jsonnet_snippet.go
hausdorff: So, the impact of this is that you have a `jsonnet.Parse` for snippets as well as a `jsonnet.Parse` for the Jsonnet language, which is a little odd, but we can fix that when we fix `Parse` to emit a `SpecificationSchema`. (We'll likely just rename it `ParseSnippet`.)

Issue #103, opened 2017-09-02
`--namespace` flag seems to be broken


hausdorff:
Following up from PR #100, which shows a failure of the integration tests, I believe that the `-n` flag is failing to be handled correctly. I was able to reproduce locally. Observe the following example, where I attempt to use `update -n` to put a `configMap` with the name `testcm` into a namespace, but when I run `get` we see that it shows up in the namespace specified in the current context:

```shell
$ ../kubecfg update -vv -n updatewtlhj -f components/kubecfg-cm.yaml
INFO  Updating configmaps testcf
DEBUG Chose API 'configmaps' for /v1, Kind=ConfigMap
DEBUG Fetching client for &APIResource{Name:configmaps,Namespaced:true,Kind:ConfigMap,Verbs:[create delete deletecollection get list patch update watch],ShortNames:[cm],} namespace=kubecfgtest
DEBUG Patch(testcf) returned (&{map[kind:ConfigMap apiVersion:v1 metadata:map[resourceVersion:3471 creationTimestamp:2017-09-02T22:30:41Z name:testcf namespace:kubecfgtest selfLink:/api/v1/namespaces/kubecfgtest/configmaps/testcf uid:5a271989-902e-11e7-875d-06e36f41fda2]]}, <nil>)
DEBUG Updated object: {"apiVersion":"v1","

A: data":{},"kind":"ConfigMap","metadata":{"name":"testcf"}}

B: kind":"ConfigMap","metadata":{"creationTimestamp":"2017-09-02T22:30:41Z","name":"testcf","namespace":"kubecfgtest","resourceVersion":"3471","selfLink":"/api/v1/namespaces/kubecfgtest/configmaps/testcf","uid":"5a271989-902e-11e7-875d-06e36f41fda2"}}
$ k get cm -n updatewtlhj
No resources found.
$ k get cm --all-namespaces
NAMESPACE     NAME                                 DATA      AGE
kube-public   cluster-info                         2         1h
kube-system   calico-config                        3         1h
kube-system   extension-apiserver-authentication   6         1h
kube-system   kube-proxy                           1         1h
kubecfgtest   nons                                 0         44m
kubecfgtest   testcf                               0         1h
kubecfgtest   testcm                               1         44m
```

hausdorff:
[...]

PR #111, opened 2017-09-05
WIP: Add Dockerfile

hausdorff:


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+#     --workdir /wd                 \
+#     ksonnet/ksonnet               \
+#     kubecfg "$@"
+# }
+
+FROM golang:1.8 as kubecfg-builder
+# Keep this in sync with the corresponding ENV in stage 2
+ENV KUBECFG_VERSION v0.5.0
+
+RUN go get github.com/anguslees/kubecfg-1
```
anguslees: heh, I'm flattered but there's no way this should be pulling from my personal fork :P

hausdorff:
To recap, I couldn't figure out how to start CI without issuing a PR, and so when we were hunting down #103, I used this to run some tests in Docker, so that I could determine it wasn't just my machine.

But, now that we've written it, we might as well just polish it up and commit it.

@anguslees what part of this do you want to be exercised by CI? I thought Travis could not build docker images.

hausdorff:
I've added a bunch of goodies to make it easy to deal with the Dockerfile, and I'm going to try to see if we can build from inside Travis to see if it can be exercised by CI.

mkmik:
This PR is stale; probably it makes more sense to add kubecfg to the bitnami application catalog so we don't need to worry with container delivery pipeline here

PR #112, opened 2017-09-06
Allow explicit --namespace to override value in kubeconfig context

### LINKED ISSUE #103: `--namespace` flag seems to be broken

anguslees:
If the current kubeconfig context is declared with an
explicit (non-empty) namespace, then `mergedKubeConfig.Namespace()`
returns that *config file* namespace value, ignoring an explicit
`--namespace` command line flag.

See https://github.com/kubernetes/client-go/issues/288 for upstream
client-go bug/discussion.

This change works around the issue by just checking the Namespace
override ourselves before falling through to the regular function.

Further fixes #103

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -50,7 +50,7 @@ var deleteCmd = &cobra.Command{
 			return err
 		}
 
-		c.DefaultNamespace, _, err = clientConfig.Namespace()
+		c.DefaultNamespace, err = defaultNamespace(clientConfig)
 		if err != nil {
 			return err
 		}
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -48,7 +48,7 @@ var diffCmd = &cobra.Command{
 			return err
 		}
 
-		c.DefaultNamespace, _, err = clientConfig.Namespace()
+		c.DefaultNamespace, err = defaultNamespace(clientConfig)
 		if err != nil {
 			return err
 		}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -60,6 +60,7 @@ const (
 )
 
 var clientConfig clientcmd.ClientConfig
+var overrides clientcmd.ConfigOverrides
 
 func init() {
 	RootCmd.PersistentFlags().CountP(flagVerbose, "v", "Increase verbosity. May be given multiple times.")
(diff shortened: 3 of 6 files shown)


anguslees:
Fwiw, I wrote the test before writing the fix - so I'm confident that it reproduces the issue without the accompanying patch.  (I miss aegis' bugfix workflow which would automatically verify that for you)

PR #115, opened 2017-09-07
Change directory structure generated by 'init' to support envs

jessicayuen:
This commit will remove the vendor/lib and vendor/schema directories
generated by 'init'. An 'environments' directory will be created from
the root app directory, with the default environment directory 'dev'
and it's containing contents rooted at 'environments'.

The intention of environments are to represent the deployment
environments of Kubernete clusters, with the following example
directory structure:

```
app-name/
  ..
  envs/
    dev/
    us-west/
      prod/
      staging/
```

### DIFF
--- a/metadata/manager.go
+++ b/metadata/manager.go
@@ -18,29 +18,29 @@ func appendToAbsPath(originalPath AbsPath, toAppend ...string) AbsPath {
 }
 
 const (
-	defaultEnvName = "dev"
+	ksonnetDir      = ".ksonnet"
+	libDir          = "lib"
+	componentsDir   = "components"
+	environmentsDir = "environments"
+	vendorDir       = "vendor"
 
-	ksonnetDir    = ".ksonnet"
-	libDir        = "lib"
-	componentsDir = "components"
-	vendorDir     = "vendor"
-	schemaDir     = "vendor/schema"
-	vendorLibDir  = "vendor/lib"
+	defaultEnvName = "default"
 
+	// Environment-specific files
 	schemaFilename         = "swagger.json"
 	ksonnetLibCoreFilename = "k8s.libsonnet"
+	specFilename           = "spec.json"
 )
 
 type manager struct {
 	appFS afero.Fs
 
-	rootPath       AbsPath
-	ksonnetPath    AbsPath
-	libPath        AbsPath
-	componentsPath AbsPath
-	vendorDir      AbsPath
-	schemaDir      AbsPath
-	vendorLibDir   AbsPath
+	rootPath        AbsPath
+	ksonnetPath     AbsPath
+	libPath         AbsPath
+	componentsPath  AbsPath
+	environmentsDir AbsPath
+	vendorDir       AbsPath
 }
 
 func findManager(abs AbsPath, appFS afero.Fs) (*manager, error) {
@@ -66,13 +66,6 @@ func findManager(abs AbsPath, appFS afero.Fs) (*manager, error) {
 }
 
 func initManager(rootPath AbsPath, spec ClusterSpec, appFS afero.Fs) (*manager, error) {
-	//
-	// IMPLEMENTATION NOTE: We get the cluster specification and generate
-	// ksonnet-lib before initializing the directory structure so that failure of
-	// either (e.g., GET'ing the spec from a live cluster returns 404) does not
-	// result in a partially-initialized directory structure.
-	//
-
 	// Get cluster specification data, possibly from the network.
 	specData, err := spec.data()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


metadata/manager.go
```diff
@@ -144,19 +130,32 @@ func (m *manager) ComponentPaths() (AbsPaths, error) {
 	return paths, nil
 }
 
-func (m *manager) LibPaths() (libPath, vendorLibPath AbsPath) {
-	return m.libPath, m.vendorLibDir
+func (m *manager) LibPaths(env string) (libPath, envLibPath AbsPath) {
+	return m.libPath, AbsPath(path.Join(environmentsDir, env))
 }
 
 func (m *manager) cacheClusterSpecData(name string, specData []byte) error {
```
jessicayuen: This definition will be changed to `func (m *manager) CreateEnvironment(name, uri string, spec ClusterSpec) error {` once `env add` is complete.

metadata/manager.go
```diff
@@ -18,29 +18,29 @@ func appendToAbsPath(originalPath AbsPath, toAppend ...string) AbsPath {
...
+	environmentsDir = "environments"
+	vendorDir       = "vendor"
 
-	ksonnetDir    = ".ksonnet"
-	libDir        = "lib"
-	componentsDir = "components"
-	vendorDir     = "vendor"
-	schemaDir     = "vendor/schema"
-	vendorLibDir  = "vendor/lib"
+	defaultEnvName = "dev"
```
hausdorff: Now that we're implementing environments, we should make a note to take a more principled cut at this. Right now we need to decide:

* What should the default environment be called? `dev` doesn't seem quite right.
* When should `apply` default to `kubeconfig`? What should we do when we `apply <env-name>` and `env-name` doesn't have a URL associated with it? (Seems like we should error out and suggest they run some command to "adopt" the current context in the currently-active `kubeconfig` file. Also whenever we run `apply`, we should probably say exactly what we're deploying to and where we got the data.)

metadata/manager.go
```diff
@@ -81,13 +81,6 @@ func initManager(rootPath AbsPath, spec ClusterSpec, appFS afero.Fs) (*manager,
 
 	m := newManager(rootPath, appFS)
 
-	// Generate the program text for ksonnet-lib.
-	ksonnetLibDir := appendToAbsPath(m.schemaDir, defaultEnvName)
-	ksonnetLibData, err := generateKsonnetLibData(ksonnetLibDir, specData)
```
hausdorff: This code can either make a remote call to the network (if initializing against a cluster) or the FS (if initializing from file), so I believe we want to do that _before_ we init the directory structure. Otherwise, if the network call fails, you get a half-initialized app, which seems bad.

Likewise, I believe we want to write out the ksonnet-lib file after we create the dir tree.

If you agree, maybe we should move the `IMPLEMENTATION NOTE` comment down to this line to make it clearer.

metadata/manager.go
```diff
@@ -144,19 +130,32 @@ func (m *manager) ComponentPaths() (AbsPaths, error) {
 	return paths, nil
 }
 
-func (m *manager) LibPaths() (libPath, vendorLibPath AbsPath) {
-	return m.libPath, m.vendorLibDir
+func (m *manager) LibPaths(env string) (libPath, envLibPath AbsPath) {
+	return m.libPath, AbsPath(path.Join(environmentsDir, env))
```
hausdorff: Instead of `AbsPath(path.Join(environmentsDir, env))`, what do you think of something like `appendToAbsPath(m.environmentsDir, env)`. Also what do you think about calling it `envName` instead of `env`?

metadata/manager.go
```diff
@@ -144,19 +130,32 @@ func (m *manager) ComponentPaths() (AbsPaths, error) {
...
 
-	schemaPath := string(filepath.Join(envPath, schemaFilename))
+	// Generate the schema file.
+	schemaPath := string(filepath.Join(string(envPath), schemaFilename))
 	err = afero.WriteFile(m.appFS, schemaPath, specData, os.ModePerm)
+	if err != nil {
+		return err
+	}
+
+	// Generate the program text for ksonnet-lib.
```
hausdorff: See comment above about the ksonnet-lib gen code going in `initManager`. I agree that it really _should_ go here, but the error handling semantics makes me think we should put it there.

hausdorff:
lol Jess accidentally deleted one of the comments. Here it was:

Now that we're implementing environments, we should make a note to take a more principled cut at this. Right now we need to decide:

* What should the default environment be called? dev doesn't seem quite right. Possibly `default`.
* When should `apply` default to using the cluster in `kubeconfig`? What should we do when we `apply <env-name>` and `env-name` doesn't have a URL associated with it? (Seems like we should error out and suggest they run some command to "adopt" the current context in the currently-active kubeconfig file. Also whenever we run apply, we should probably say exactly what we're deploying to and where we got the data.)

metadata/manager.go
hausdorff: Sounds good!

metadata/manager.go
```diff
@@ -144,19 +136,26 @@ func (m *manager) ComponentPaths() (AbsPaths, error) {
...
+func (m *manager) cacheClusterSpecData(name string, specData, ksonnetLibData []byte) error {
+	envPath := appendToAbsPath(m.environmentsDir, name)
+	err := m.appFS.MkdirAll(string(envPath), os.ModePerm)
 	if err != nil {
 		return err
 	}
 
-	schemaPath := string(filepath.Join(envPath, schemaFilename))
+	// Generate the schema file.
+	schemaPath := string(filepath.Join(string(envPath), schemaFilename))
```
hausdorff: Just 2 suggestions here:

* How about something like `schemaPath := appendToAbsPath(envPath, schemaFilename)` instead of the above, and then you do the `string` conversion as an argument to `WriteFile`?
* What do you think about `envName` instead of `name`?

hausdorff:
@anguslees @jessicayuen: Just FYI, I'm not sure what's going on here, but when I merged this, GitHub claimed (and still claims, above) that the build passed. But on commits page for `master`, it displays [this error](https://travis-ci.org/ksonnet/kubecfg/builds/273492800?utm_source=github_status&utm_medium=notification) on the merge commit.

Travis CI's SAN is also [completely incapacitated, with no ETA](https://twitter.com/traviscistatus/status/906268368897957889), so that may or may not be related. Either way, I can't check what's going on until it comes back.

PR #119, opened 2017-09-08
'init' should use current directory when generating Ksonnet app

jessicayuen:


hausdorff:
Travis CI [is broken](https://twitter.com/traviscistatus/status/906268368897957889). I'll merge this whenever it comes back (they have "no ETA" for when it will come back, and that update was 2 hours ago, so it might be several more hours).

PR #121, opened 2017-09-10
Fix #113

### LINKED ISSUE #113: `kubecfg init` does not include k.libsonnet
k8.libsonnet (in "/vendor/schema/dev /") is only created with 'kubecfg init' should also include 'k.libsonnet' somewhere, maybe the whole ksonnet.beta.2 folder (?)

hausdorff:


anguslees:
Please add a description that gives some idea of the change.  A future me that is doing some complicated git history search thanks you.

vendor/github.com/ksonnet/ksonnet-lib/ksonnet-gen/ksonnet/emit.go
```diff
@@ -13,12 +13,19 @@ import (
 
 // Emit takes a swagger API specification, and returns the text of
 // `ksonnet-lib`, written in Jsonnet.
```
jessicayuen: Not directly related to this PR, but I was trying to make sense of the dependency changes ... would have been nice to see this comment updated to reflect the new return objects.

metadata/manager.go
```diff
@@ -184,12 +191,12 @@ func (m *manager) createAppDirTree() error {
 	return nil
 }
 
-func generateKsonnetLibData(ksonnetLibDir AbsPath, text []byte) ([]byte, error) {
+func generateKsonnetLibData(ksonnetLibDir AbsPath, text []byte) ([]byte, []byte, error) {
```
jessicayuen: nil is always being returned for the second return object it seems

hausdorff:
Added more detailed commit message per @anguslees. Will merge when CI passes.

PR #122, opened 2017-09-10
Add types for prototype parameters

### LINKED ISSUE #116: Implement typing for Jsonnet parameters
When we `use` prototypes that emit Jsonnet, it's often unclear how a parameter should be emitted. For example:

```
$ ks prototype use --replicas 3
```

It is conceivable that `3` could be a string or a number; in YAML this doesn't matter, but in Jsonnet it does.

This change consists mostly of altering `prototype.ParamSchema` needs to contain a `Type` member, and then making sure we use it when we emit Jsonnet-type prototype templates.

hausdorff:
This commit fixes #116 by introducing two new constructs to the
prototype specification schema:

1. Mandatory types for prototype parameters. This lets us accept bare
   words on the command line, and then "do the right thing" when
   emitting JSON or Jsonnet.

   For example, say a template produces a `core.v1.Service` that exposes
   a port with a `--targetPort` flag. When the user passes a number
   (e.g., `80`) in, we should _not_ put quote marks around it, since we
   want to expose port `80`. When the user passes a string (e.g.,
   `"nginxPort"`), we _should_ put quote marks around it, to denote that
   we're exposing the port with that name.

   In order to do this, we need to know they "type" of the parameter (in
   this case, `NumberOrString`).

2. Mandatory template types. A template can have a JSON, YAML, or
   Jsonent flavor, and we default to using Jsonnet. This is useful
   mostly to make type parameters less error-prone (since one set of
   parameters corresponds to one set of templates), but it also
   significantly de-bloats the output of commands like `search`, since
   one fully-qualified name can correspond to multiple flavors of the
   same template.


prototype/specification.go
```diff
@@ -26,6 +27,52 @@ type SpecificationSchema struct {
...
+			maxNameDescLen = nameDescLen
+		}
+	}
+
+	lines := []string{}
+	for _, proto := range ss {
+		// NOTE: If we don't add 1 below, the longest name will look like :
+		// `io.whatever.pkg.fooDescription is here.`
+		nameSpace := strings.Repeat(" ", maxNameLen-len(proto.Name)+1)
+		descSpace := strings.Repeat(" ", maxNameDescLen-maxNameLen-len(proto.Template.ShortDescription)+2)
```
jessicayuen: Do 'ShortDescription's have a line length constraint on them? If not, should a line length be enforced for displaying (by trimming or otherwise) to make sure formatting appears ok?

hausdorff: This is a good suggestion, but I wasn't sure so I did the simple thing. I considered wrapping the description, but I'm not sure that's better.

Issue #99, opened 2017-08-31
show_test.go's TestShow failing intermittently


jessicayuen:
@hausdorff and I have noticed that this tests fails occasionally.

```
go test -ldflags="-X main.version=build-270159619 -linkmode external -extldflags=-static"  ./cmd/... ./utils/... ./pkg/... ./metadata/...
--- FAIL: TestShow (0.24s)
	show_test.go:34: Running args [show -J ../testdata/lib -o yaml -f ../testdata/test.jsonnet -V aVar=aVal -V anVar --ext-str-file filevar=../testdata/extvar.file]
	show_test.go:90: output is ---
		apiVersion: v0alpha1
		array:
		- one
		- 2
		- - 3
		bool: true
		filevar: |
		  foo
		kind: TestObject
		nil: null
		notAVal: aVal
		notAnotherVal: aVal2
		number: 42
		object:
		  foo: bar
		string: bar
		
	show_test.go:34: Running args [show -J ../testdata/lib -o json -f ../testdata/test.jsonnet -V aVar=aVal -V anVar --ext-str-file filevar=../testdata/extvar.file]
	show_test.go:90: output is ---
		{
		  "apiVersion": "v0alpha1",
		  "array": [
		    "one",
		    2,
		    [
		      3
		    ]
		  ],
		  "bool": true,
		  "filevar": "foo\n",
		  "kind": "TestObject",
		  "nil": null,
		  "notAVal": "aVal",
		  "notAnotherVal": "aVal2",
		  "number": 42,
		  "object": {
		    "foo": "bar"
		  },
		  "string": "bar"
		}
		---
		{
		  "apiVersion": "v0alpha1",
		  "array": [
		    "one",
		    2,
		    [
		      3
		    ]
		  ],
		  "bool": true,
		  "filevar": "foo\n",
		  "kind": "TestObject",
		  "nil": null,
		  "notAVal": "aVal",
		  "notAnotherVal": "aVal2",
		  "number": 42,
		  "object": {
		    "foo": "bar"
		  },
		  "string": "bar"
		}
		
	show_test.go:93: error parsing output of format json: invalid character '-' in numeric literal
FAIL
FAIL	github.com/ksonnet/kubecfg/cmd	0.350s
ok  	github.com/ksonnet/kubecfg/utils	0.726s
ok  	github.com/ksonnet/kubecfg/pkg/kubecfg	0.115s
ok  	github.com/ksonnet/kubecfg/metadata	0.018s
[...]

PR #124, opened 2017-09-11
Expose basic cluster information to jsonnet

anguslees:
This change adds two new jsonnet native functions:

- `kubernetesVersion` returns (eg) `[1, 7]` for a K8s 1.7 server.
- `kubernetesGroupVersionSupported` returns a bool indicating whether
  a given GroupVersion is supported by the current server

Note the latter does not provide a way to iterate over all supported
GroupVersions, for privacy.


anguslees:
Discuss.  In particular:

- I'm not crazy about the length of the name `kubernetesGroupVersionSupported`

- I made `kubernetesVersion` return [int, int] (rather than the native [str, str] which k8s exposes) since I thought doing less-than/greater-than comparisons would be common.  The problem with this is that it's _actually_ [float, float] because jsonnet inherits json's types - and I'm mildly concerned this is theoretically not comparable for equality (also a common expected operation).

anguslees:
oh, and the first of the two commits rewrites `show_test.go` as an integration test.  Fixes #99

PR #130, opened 2017-09-12
[docs] Add autogenerated CLI docs

abiogenesis-now:
@hausdorff PTAL---I figured it might be useful to have autogenerated markdown docs for the CLI that users can peruse in addition to using `--help` in the CLI itself. Let me know if you think this is unnecessary though.

**NOTE**: I'm going to do a second pass at some point today/tomorrow to actually vet the CLI helper text, but wanted to get this PR in separately.

This is pretty similar to what I did previously with Ark, with some minor modifications to make it work.


abiogenesis-now:
Porting this over to the ksonnet/ksonnet repo

PR #131, opened 2017-09-12
Introduce env subcommands 'env add', 'env rm', 'env set' & 'env list'

jessicayuen:
```
'env add <env-name> <env-uri>' will create a new environment within a
ksonnet project, by generating a new directory, 'env-name', within the
'envs' directory. Each environment will contain environment-specfic
files. Notably, a new environment-specific file is 'spec.json'.
'spec.json' currently only contains the 'env-uri' of the Kubernetes
cluster located at the added environment.

Below is an example directory structure for the environment
'us-west/staging':

app-name/
  .gitignore           Default .gitignore; can customize VCS
  .ksonnet/            Metadata for ksonnet
  environments/        Env specs (defaults: dev, test, prod)
    default/           [Default generated environment.]
    us-west/           [Example of user-generated env]
      staging/
        k.libsonnet
        k8s.libsonnet
        swagger.json
        spec.json      [This will contain the uri of the environment]
  components/          Top-level Kubernetes objects defining application
  lib/                 user-written .libsonnet files
  vendor/              mixin libraries, prototypes
```

```
'env list' will list all environments within a Ksonnet project. Each
environment will be pretty-printed with it's name and cluster URI
location.
```

```
'env rm <env-name>' deletes an environment within a ksonnet project.
This is the same as removing the <env-name> environment directory and
all files contained.  Empty parent directories will also be removed.
```

```
'env set <name>' sets environment fields such as the name, and cluster
URI. It currently accepts the flags '--name' and '--uri'. Changing the
name of an environment will also update the directory structure in
'environments'.
```


cmd/env.go
```diff
@@ -0,0 +1,234 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		flags := cmd.Flags()
+		if len(args) != 2 {
+			return fmt.Errorf("'env add' takes two arguments, the name and the uri of the environment, respectively")
+		}
+
+		envName := args[0]
+		envURI := args[1]
+
+		appDir, err := os.Getwd()
```
hausdorff: Non-blocking comment, I notice that we're just using `os.Getwd` and just casting that to `AbsPath` in every command, and we may or may not want to rewrite all these uses to fix this.

cmd/env.go
```diff
@@ -0,0 +1,234 @@
...
+		if err != nil {
+			return err
+		}
+		appRoot := metadata.AbsPath(appDir)
+
+		specFlag, err := flags.GetString(flagAPISpec)
```
hausdorff: Tiny comment: might want to move `flagAPISpec` into `root.go` and out of `init.go` since now we're using it in between a bunch of commands.

cmd/env.go
```diff
@@ -0,0 +1,234 @@
...
+		  us-west/           [Example of user-generated env]
+			staging/
+			  k.libsonnet
+			  k8s.libsonnet
+			  swagger.json
+			  spec.json      [This will contain the uri of the environment]
+		components/          Top-level Kubernetes objects defining application
+		lib/                 user-written .libsonnet files
+		vendor/              mixin libraries, prototypes
+`,
```
hausdorff: I don't know how we want to write these docs -- I have some particular ideas, so are you open to me rewriting them to fit what I'm looking for?

cmd/env.go
```diff
@@ -0,0 +1,234 @@
...
+			return err
+		}
+		appRoot := metadata.AbsPath(appDir)
+
+		c, err := kubecfg.NewEnvListCmd(appRoot)
+		if err != nil {
+			return err
+		}
+
+		formattedEnvsString, err := c.Run()
```
hausdorff: Instead of returning a list of strings, what do you think about doing something like `show` does:

```
return c.Run(objs, cmd.OutOrStdout())
```

cmd/env.go
```diff
@@ -0,0 +1,234 @@
...
+		fmt.Print(formattedEnvsString)
+
+		return nil
+	},
+	Long: `List all environments within a ksonnet project. This will
+display the name and the URI of each environment within the ksonnet project.`,
+}
+
+var envSetCmd = &cobra.Command{
+	Use:   "set <env-name> [parameter-flags]",
```
hausdorff: I know we discussed not allowing for renaming, and sorry for flip-flopping, but now that I think about it, I think that it would be useful to have a `kubectl`-style `--name` flag.

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+const (
+	defaultEnvName = "default"
+
+	schemaFilename        = "swagger.json"
+	extensionsLibFilename = "k.libsonnet"
+	k8sLibFilename        = "k8s.libsonnet"
+	specFilename          = "spec.json"
+)
+
+// Environment represents all fields of a Ksonnet environment
```
hausdorff: Also, extremely micro thing, but marketing has determined that ksonnet use a lower-case k. :)

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+	Name string
+	URI  string
+}
+
+// EnvironmentSpec represents the contents in spec.json.
+type EnvironmentSpec struct {
+	URI string `json:"uri"`
+}
+
+func (m *manager) CreateEnvironment(name, uri string, spec ClusterSpec, extensionsLibData, k8sLibData []byte) error {
```
hausdorff: What do you think about putting all the `manager` functions in `manager.go`?

metadata/interface.go
```diff
@@ -21,14 +36,17 @@ type Manager interface {
 	Root() AbsPath
 	ComponentPaths() (AbsPaths, error)
 	LibPaths(envName string) (libPath, envLibPath AbsPath)
+	GenerateKsonnetLibData(spec ClusterSpec) ([]byte, []byte, error)
```
hausdorff: Hmm, what do you think about making `GenerateKsonnetLibData` a private member? It doesn't look like we _need_ to have the results available in the `cmd/` package. It seems like it could be hidden behind `CreateEnvironment`. The call to `EnvAddCmd.Run` becomes simpler, and `CreateEnvironment` doesn't have to have so many parameters. And, then you don't have to get `spec.data` multiple times (as we do in, _e.g._, `env add`).

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+	envPath := appendToAbsPath(m.environmentsDir, name)
+	err := m.appFS.MkdirAll(string(envPath), os.ModePerm)
+	if err != nil {
+		return err
+	}
+
+	// Get cluster specification data, possibly from the network.
+	specData, err := spec.data()
```
hausdorff: I mention this below also, but we're calling `spec.data()` twice, right? Once in `manager.GenerateKsonnetLibData` and once here. Since this involves a possible network command, we should probably try to do it once. (See comments about `GenerateKsonnetLibData` in `interface.go` below.)

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// EnvironmentSpec represents the contents in spec.json.
+type EnvironmentSpec struct {
+	URI string `json:"uri"`
+}
+
+func (m *manager) CreateEnvironment(name, uri string, spec ClusterSpec, extensionsLibData, k8sLibData []byte) error {
+	envPath := appendToAbsPath(m.environmentsDir, name)
+	err := m.appFS.MkdirAll(string(envPath), os.ModePerm)
```
hausdorff: I propose that we error out here if the env already exists.

Also, what happens if there's `..` in the path somewhere? Even if we assume environments are not nested, It seems like we should handle that.

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+func (m *manager) DeleteEnvironment(name string) error {
+	envPath := string(appendToAbsPath(m.environmentsDir, name))
+
+	// Check whether this environment exists
+	envExists, err := m.environmentExists(name)
+	if err != nil {
+		return err
+	}
+	if !envExists {
+		return fmt.Errorf("Environment \"%s\" does not exist", name)
```
hausdorff: Tiny tiny thing, elsewhere we use `'` instead of `"` for this sort of thing. Here and below.

metadata/environment.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	// Remove the directory and all files within the environment path.
+	err = m.appFS.RemoveAll(envPath)
+	if err != nil {
+		return err
+	}
+
+	// Need to ensure empty parent directories are also removed.
+	dirs := strings.Split(name, "/")
```
hausdorff: So, I think it's sort of dangerous to split on `/`. I think we should do something more like what we're doing in `findManager`, where we call `filepath.Dir` repeatedly. This way we let the standard library handle the problem of chopping off the last component.

cmd/env.go
jessicayuen: Oh I can see how this is confusing. It does have a --name flag, but I still need to know the existing name <env-name> to know what to change it to.

cmd/env.go
jessicayuen: nope!

metadata/environment.go
jessicayuen: I felt manager.go was getting crowded. Having environment.go and environment_test.go made it easier to distinguish what was happening with environments. But that said, I'm not sure what go best practices are with multiple files sharing an interface implementation. Do you have a strong preference on this?

hausdorff: Alright, that's fine. We can keep them in their own file.

hausdorff:
Alright, a couple proposals, which we should write issues for if we agree:

* Now that we have the mechanics down, we probably want to consider more carefully what we _really_ want to happen during `init`. We likely want as close to a one-command run to get a fully-functional environment. Right now, the if the kubconfig file can't be found, we will just generate a `spec.json` that's got an empty URI. I think we might want to:
  1. When we run `init`, output what we're generating `default` from. Right now, I believe it's nothing unless you explicitly specify the URI. (_e.g._, we might output something like `Generating 'default' environment from current context in kubeconfig file at '~/whatever'`.)
  1.  I think if we want to get the URI from the current context, it's probably fine to just print out where we got it from.
  1. If there is no kubeconfig file, don't generate a default environment, and explain why. (_e.g._, `Could not generate 'default' environment, kubeconfig file not found`)
* `env list` should output more information than just the name (_i.e._, perhaps the URI or the way we're ID'ing the cluster.
* We should get formal about how the `env` commands produce a unique ID for a cluster. I'm thinking a hash of URI + server cert? This is a temporary solution of course, we'll need to fix it up stream.
* `spec.json` (and `swagger.json`, now that I think of it) is marked as executable, that's probably not what we want.

jessicayuen:
> When we run init, output what we're generating default from. Right now, I believe it's nothing unless you explicitly specify the URI. (e.g., we might output something like Generating 'default' environment from current context in kubeconfig file at '~/whatever'.)

Agreed. The default environment right now isn't generated from kubeconfig. It's simply an empty URI and this case needs to be handled soon.  Probably also need to consider in this case whether we want users to use `ksonnet apply default`, or implement a special case `ksonnet apply` that will just look for the default environment if <env> is not specified.

> env list should output more information than just the name (i.e., perhaps the URI or the way we're ID'ing the cluster.

It currently outputs the name and the URI. I think you're just seeing the name because of how the default env has an empty URI currently.

> We should get formal about how the env commands produce a unique ID for a cluster. I'm thinking a hash of URI + server cert? This is a temporary solution of course, we'll need to fix it up stream.

Agreed.

> spec.json (and swagger.json, now that I think of it) is marked as executable, that's probably not what we want.

I'll open a bug for this.

jessicayuen:
Opened the following: #134, #135, #136, #137

hausdorff:
Also:

* Totally minor point, but we might want to allow the identity renaming, _e.g._, `ks env set foo --name foo`. It just seems a little odd to error out here, since it seems to indicate something that's technically valid.
* The `env set` command I believe will allow you to do something like `env set foo --name ../foo`
* If users don't specify a URI argument to `env add`, _i.e._, `ks env add foo`, we probably want to just pull from the current context -- and as with `init`, we probably want to write it down when we do.

jessicayuen:
> Totally minor point, but we might want to allow the identity renaming, e.g., ks env set foo --name foo. It just seems a little odd to error out here, since it seems to indicate something that's technically valid.

I'll add the fix in this PR.

> The env set command I believe will allow you to do something like env set foo --name ../foo

I'll add the fix in this PR.

> If users don't specify a URI argument to env add, i.e., ks env add foo, we probably want to just pull from the current context -- and as with init, we probably want to write it down when we do.

I'll open an `enhancement` issue for this.

cmd/root.go
```diff
@@ -70,6 +71,8 @@ func init() {
 	RootCmd.PersistentFlags().StringSlice(flagTlaVarFile, nil, "Read top level argument from a file")
 	RootCmd.PersistentFlags().String(flagResolver, "noop", "Change implementation of resolveImage native function. One of: noop, registry")
 	RootCmd.PersistentFlags().String(flagResolvFail, "warn", "Action when resolveImage fails. One of ignore,warn,error")
+	// TODO: We need to make this default to checking the `kubeconfig` file.
+	RootCmd.PersistentFlags().String(flagAPISpec, "version:v1.7.0", "Manually specify API version from OpenAPI schema, cluster, or Kubernetes version")
```
hausdorff: I actually think we only want to move the `flagAPISpec` member. I don't think we want every command to have this flag (_i.e._, just stuff like `init`).

metadata/environment.go
```diff
@@ -0,0 +1,285 @@
...
+		desiredExists, err := m.environmentExists(desired.Name)
+		if err != nil {
+			return err
+		}
+		if desiredExists {
+			return fmt.Errorf("Can not update '%s' to '%s', it already exists", name, desired.Name)
+		}
+
+		// Move the directory
+		pathOld := string(appendToAbsPath(m.environmentsPath, name))
```
hausdorff: I think this might not handle the case like `--name ../foo`. I think we should pull validation logic into some other function, `validateName` and then run this through that.

metadata/environment.go
```diff
@@ -0,0 +1,285 @@
...
+
+		if isDir {
+			// Only want leaf directories containing a spec.json
+			specPath := filepath.Join(path, specFilename)
+			specFileExists, err := afero.Exists(m.appFS, specPath)
+			if err != nil {
+				return err
+			}
+			if specFileExists {
+				envName := filepath.Clean(strings.TrimPrefix(path, string(m.environmentsPath)+"/"))
```
hausdorff: If the user adds an env with a name that has a trailing `/` to the end of their name, this will act strangely, no? We should probably disallow trailing slashes in `CreateEnvironment` and `SetEnvironment`.

metadata/environment.go
jessicayuen: I added this handler a few minutes ago (that set & add use), if you want to double check for me again,
```
// regex matcher to ensure environment name does not contain punctuation
func isValidName(envName string) bool {
	isInvalidName := regexp.MustCompile(`[,;.':!()?"{}\[\]]+`).MatchString
	return !isInvalidName(envName)
}
```

metadata/environment.go
jessicayuen: Added the trailing slash check, see below comment

Issue #97, opened 2017-08-31
Environments not implemented


hausdorff:
Right now most commands expect an environment name. We should implement simple environments so this makes sense, or make the arg optional.

PR #140, opened 2017-09-18
Commands using the <env> arg should deploy to the correct cluster

jessicayuen:
For example, 'apply <env' currently operates as a no-op. With the
introduction of simple environments in PR #131, 'apply <env>' should
perform basic validation such that:

1. The user has added the environment that is being deployed against to
their Ksonnet project.

2. The URI in the environment's spec file that the user wishes to deploy
to should correspond to at least one cluster location as listed in
kubeconfig.

If either of those conditions are not satisfied, the kubecfg user will
receive the corresponding error.

In addition, this commit will set the kubectl --cluster flag to point at
the cluster listed by the environment URI.

Related: #97

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -299,14 +299,11 @@ func expandEnvCmdObjs(cmd *cobra.Command, args []string) ([]*unstructured.Unstru
 	envPresent := env != nil
 	filesPresent := len(fileNames) > 0
 
-	// This is equivalent to: `if !xor(envPresent, filesPresent) {`
-	if envPresent && filesPresent {
-		return nil, fmt.Errorf("Either an environment name or a file list is required, but not both")
-	} else if !envPresent && !filesPresent {
-		return nil, fmt.Errorf("Must specify either an environment or a file list")
+	if !envPresent && !filesPresent {
+		return nil, fmt.Errorf("Must specify either an environment or a file list, or both")
 	}
 
-	if envPresent {
+	if envPresent && !filesPresent {
 		manager, err := metadata.Find(metadata.AbsPath(cwd))
 		if err != nil {
 			return nil, err
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -226,7 +227,14 @@ func dumpJSON(v interface{}) string {
 	return string(buf.Bytes())
 }
 
-func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.DiscoveryInterface, error) {
+func restClientPool(cmd *cobra.Command, envName *string) (dynamic.ClientPool, discovery.DiscoveryInterface, error) {
+	if envName != nil {
+		err := overrideCluster(*envName)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
 	conf, err := clientConfig.ClientConfig()
 	if err != nil {
 		return nil, nil, err
@@ -245,6 +253,11 @@ func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.Discovery
 	return pool, discoCache, nil
 }
 
+type envSpec struct {
+	env   *string
+	files []string
+}
+
 // addEnvCmdFlags adds the flags that are common to the family of commands
 // whose form is `[<env>|-f <file-name>]`, e.g., `apply` and `delete`.
 func addEnvCmdFlags(cmd *cobra.Command) {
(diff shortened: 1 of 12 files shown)


cmd/root.go
```diff
@@ -299,14 +299,11 @@ func expandEnvCmdObjs(cmd *cobra.Command, args []string) ([]*unstructured.Unstru
...
-	if envPresent && filesPresent {
-		return nil, fmt.Errorf("Either an environment name or a file list is required, but not both")
-	} else if !envPresent && !filesPresent {
-		return nil, fmt.Errorf("Must specify either an environment or a file list")
+	if !envPresent && !filesPresent {
+		return nil, fmt.Errorf("Must specify either an environment or a file list, or both")
 	}
 
-	if envPresent {
+	if envPresent && !filesPresent {
```
hausdorff: We should probably update all the docs to reflect this non-mutual-exclusion.

jessicayuen: Updated.

cmd/root.go
```diff
@@ -253,32 +259,86 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
 // while if a user passes an environment name, we will expand all component
 // files using that environment.
-func expandEnvCmdObjs(cmd *cobra.Command, args []string) ([]*unstructured.Unstructured, error) {
-	env, fileNames, err := parseEnvCmd(cmd, args)
-	if err != nil {
-		return nil, err
-	}
+func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd string) ([]*unstructured.Unstructured, error) {
+	env := envSpec.env
+	fileNames := envSpec.files
```
hausdorff: Hmm, is there any utility to using a `fileNames` as an intermediate value here, or do you think we should just use `envSpec.files`? Here and with `env`.

metadata/environment.go
```diff
@@ -189,6 +189,21 @@ func (m *manager) GetEnvironments() ([]Environment, error) {
 	return envs, nil
 }
 
+func (m *manager) GetEnvironment(name string) (Environment, error) {
```
hausdorff: Now that I think of it, I think it's worth regrouping and deciding whether we should be returning a pointer or a value here. I should have caught this also in `SetEnvironment`. I think it should be a pointer, because that's what people are used to.

hausdorff: I'm fine moving the fix to `SetEnvironment` to another PR if we file an issue and put it in the MLP milestone.

jessicayuen: I'll fix both cases in this PR

cmd/root.go
```diff
@@ -253,33 +259,84 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
+}
+
+// overrideCluster ensures that the cluster URI specified in the environment is
+// associated in the user's kubeconfig file during deployment to a ksonnet
+// environment. We will error out if it is not.
+//
+// If the environment URI the user is attempting to deploy to is not the current
+// kubeconfig context, we must manually override the client-go --cluster flag
+// to ensure we are deploying to the correct cluster.
+func overrideCluster(env *string, cwd string) error {
```
hausdorff: Probably `cwd` should be type `AbsPath` when it comes in. We're expecting it to be an `AbsPath` anyway, so might as well make it formal.

cmd/root.go
```diff
@@ -253,33 +259,84 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
+	overrides.Context.Cluster = clusterURI
+
+	return nil
+}
+
+// validateEnvDeploy checks to ensure that the environment we are trying to
+// deploy to is created, and that the environment URI is located in kubeconfig.
+// It returns the environment URI if these checks pass.
+func validateEnvDeploy(envName string, clusterURIs map[string]string, wd metadata.AbsPath) (string, error) {
+	metadataManager, err := metadata.Find(wd)
```
hausdorff: Hmm, ideally we'd want to do this search once, rather than twice for every command.

cmd/root.go
```diff
@@ -253,33 +259,84 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
+	if err != nil {
+		return "", err
+	}
+
+	env, err := metadataManager.GetEnvironment(envName)
+	if err != nil {
+		return "", err
+	}
+
+	if _, ok := clusterURIs[env.URI]; ok {
```
hausdorff: Do we have to normalize this URI before storage or check? I'm thinking about something like there's a trailing `/` which doesn't mean anything semantically, and we store it one way but check it another way.

cmd/root.go
```diff
@@ -253,33 +259,84 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
+	log.Debugf("Overwriting --cluster flag with '%s'", clusterURI)
+	overrides.Context.Cluster = clusterURI
+
+	return nil
+}
+
+// validateEnvDeploy checks to ensure that the environment we are trying to
+// deploy to is created, and that the environment URI is located in kubeconfig.
+// It returns the environment URI if these checks pass.
+func validateEnvDeploy(envName string, clusterURIs map[string]string, wd metadata.AbsPath) (string, error) {
```
hausdorff: Since this is called once I would just make this a part of `overrideCluster`.

cmd/root.go
```diff
@@ -253,33 +259,84 @@ func addEnvCmdFlags(cmd *cobra.Command) {
...
+		clusterURIs[cluster.Server] = name
+	}
+
+	log.Debugf("Validating deployment at '%s' with cluster URIs '%v'", env, reflect.ValueOf(clusterURIs).MapKeys())
+	uri, err := validateEnvDeploy(*env, clusterURIs, metadata.AbsPath(cwd))
+	if err != nil {
+		return err
+	}
+
+	clusterURI := clusterURIs[uri]
```
hausdorff: I'm not super clear on what's going on here, I think this should be called `clusterName`?

cmd/root.go
jessicayuen: Opened #144

PR #141, opened 2017-09-19
Add to 'env' command documentation

### LINKED ISSUE #139: Fix up documentation in env commands

hausdorff:
Fixes #139.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -49,7 +49,35 @@ func init() {
 
 var envCmd = &cobra.Command{
 	Use:   "env",
-	Short: `Create, remove, modify, and list ksonnet environments`,
+	Short: `Manage ksonnet environments`,
+	Long: `An environment acts as a sort of "named cluster", allowing for commands like
+'ksonnet apply dev', which applies the ksonnet application to the "dev cluster".
+Additionally, environments allow users to cache data about the cluster it points
+to, including data needed to run 'verify', and a version of ksonnet-lib that is
+generated based on the flags the API server was started with (e.g., RBAC enabled
+or not).
+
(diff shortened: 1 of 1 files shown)


cmd/env.go
```diff
@@ -85,27 +85,41 @@ var envAddCmd = &cobra.Command{
...
+An environment contains no user-specific data (such as the private key
+often contained in a kubeconfig file), and
+
+Environments are represented as a hierarchy in the 'environments' directory of a
+ksonnet application. For example, in the example below, there are two
+environments: 'default' and 'us-west/staging'. Each contains a cached version of
+ksonnet-lib, and a 'spec.json' that contains the URI and server cert that
+uniquely identifies the cluster.
+
+environments/        Env specs (defaults: dev, test, prod)
```
jessicayuen: We should update the `(defaults: dev, test, prod)` part since it's no longer true. I think I missed this.

cmd/env.go
```diff
@@ -85,27 +85,41 @@ var envAddCmd = &cobra.Command{
...
-			  k.libsonnet
-			  k8s.libsonnet
-			  swagger.json
-			  spec.json      [This will contain the uri of the environment]
-		components/          Top-level Kubernetes objects defining application
-		lib/                 user-written .libsonnet files
-		vendor/              mixin libraries, prototypes
+	Long: `Add a new environment to a ksonnet project.
+
+An environment acts as a sort of "named cluster", allowing for commands like
```
jessicayuen: We should probably also add (or move) these few paragraphs to the general `env` command, where a subcommand hasn't been provided. I think it will be helpful to those that have no clue about envs and do `ksonnet env --help`

PR #145, opened 2017-09-20
Adopt 0755 as default permissions for folders

hausdorff:
Under certain circumstances, it is possible for operations to fail
because of permissions that are too restrictive. Here we adopt the
default permissions of 0755 (i.e., rwxr-xr-x), which should be a better
balance of restriction and flexibility.


hausdorff:
I think so. This is actually more permissive.

hausdorff:
(I don't think Jess can approve again, so I got a verbal approve with the new changes.

PR #149, opened 2017-09-21
Cause `init` to generate env using current-context

hausdorff:
When we run `init`, currently we generate a simple environment called
'default' with no URI. A better idea is to generate the URI from the
current context of the active kubeconfig file, if it exists.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -20,6 +20,8 @@ import (
 	"os"
 	"path"
 
+	"k8s.io/client-go/tools/clientcmd/api"
+
 	"github.com/ksonnet/kubecfg/metadata"
 	"github.com/ksonnet/kubecfg/pkg/kubecfg"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


cmd/init.go
```diff
@@ -53,7 +55,37 @@ var initCmd = &cobra.Command{
...
+				currCtx = ctx
+			}
+		}
+
+		var currClusterURI *string
+		if rawConfig.CurrentContext != "" && currCtx != nil {
+			var currCluster *api.Cluster
+			for name, cluster := range rawConfig.Clusters {
+				if currCtx.Cluster == name {
+					currCluster = cluster
```
jessicayuen: Why not just set `currClusterURI` here instead of having another variable to track `currCluster`?

hausdorff: lol good idea

hausdorff: lol good idea

PR #150, opened 2017-09-21
Remove client-go flags from commands that don't use

hausdorff:
DON'T MERGE. This is meant to test whether the client-go flags can be removed from commands that don't use them.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -50,6 +50,8 @@ func init() {
 	RootCmd.AddCommand(applyCmd)
 
 	addEnvCmdFlags(applyCmd)
+	bindClientGoFlags(applyCmd)
+	bindJsonnetFlags(applyCmd)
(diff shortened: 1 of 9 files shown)


cmd/apply.go
```diff
@@ -50,6 +50,8 @@ func init() {
 	RootCmd.AddCommand(applyCmd)
 
 	addEnvCmdFlags(applyCmd)
+	bindClientGoFlags(applyCmd)
```
anguslees: you've called these new functions `bind*Flags()` and the older similar functions `add*Flags()`.  `bindClientGoFlags()` *does* actually bind flags to other structures, so that name makes sense.  I think `bindJsonnetFlags()` should be renamed to `addJsonnetFlags()` for consistency, however.

hausdorff:
Per offline conversation with @anguslees, we're going to merge this today.

PR #151, opened 2017-09-21
Add `prototype list` command

hausdorff:
The `prototype list` command lists all known prototypes. It is
functionally equivalent to a `prototype search` that returns all
prototypes.

### DIFF
--- a/prototype/index.go
+++ b/prototype/index.go
@@ -13,6 +13,14 @@ type index struct {
 	prototypes map[string]*SpecificationSchema
 }
 
+func (idx *index) List() (SpecificationSchemas, error) {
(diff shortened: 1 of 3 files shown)


prototype/index.go
```diff
@@ -13,6 +13,14 @@ type index struct {
 	prototypes map[string]*SpecificationSchema
 }
 
+func (idx *index) List() (SpecificationSchemas, error) {
```
jessicayuen: What are your thoughts on dropping `error` from the return since there is no chance of error in the body?

hausdorff: Eventually we'll need to be able to return an error when we call out to disk for other prototypes. So unfortunately I think it should stay.

PR #154, opened 2017-09-21
Implement 'prototype preview'

hausdorff:


cmd/prototype.go
```diff
@@ -203,82 +202,214 @@ var prototypeUseCmd = &cobra.Command{
...
+	Long: `Expand prototype uniquely identified by (possibly partial)
+'prototype-name', filling in parameters from flags, and emitting the generated
+code to stdout.
+
+Note also that 'prototype-name' need only contain enough of the suffix of a name
+to uniquely disambiguate it among known names. For example, 'deployment' may
+resolve ambiguously, in which case 'use' will fail, while 'deployment' might be
+unique enough to resolve to 'io.ksonnet.pkg.single-port-deployment'.`,
+
+	Example: `  # Instantiate prototype 'io.ksonnet.pkg.single-port-deployment', using the
```
jessicayuen: `Preview` might be better than `Instantiate` in these examples, since Instantiate seems better reserved for `prototype use`

metadata/interface.go
```diff
@@ -85,6 +89,21 @@ func ParseClusterSpec(specFlag string) (ClusterSpec, error) {
 	return parseClusterSpec(specFlag, appFS)
 }
 
+// regex matcher to ensure environment name does not contain punctuation
+func isValidName(envName string) bool {
```
jessicayuen: May want to update the arg name and comment since it's no longer just used for environments

metadata/interface.go
```diff
@@ -85,6 +89,21 @@ func ParseClusterSpec(specFlag string) (ClusterSpec, error) {
...
+	// No unicode whitespace is allowed. `Fields` doesn't handled trailing or
```
jessicayuen: nit -- typo `handled` > `handle`

PR #161, opened 2017-09-27
Introduce namespace to environments

### LINKED ISSUE #148: Do not use kubeconfig current namespace when using environments
Currently if you run a command like `kubecfg apply default`, it will automatically use the namespace of the current-context. This is obviously nonsensical, since the current context might have nothing to do with the cluster we're deploying to (_i.e._, it might refer to a namespace on a totally different cluster.

jessicayuen:
Environments currently have the concept of a server URI, but it is
ambiguous which cluster namespace to use.

This commit will introduce the concept of namespaces to the env
commands.

For example,
`kubecfg env add staging http://mock-staging-uri \
 --namespace=staging-namespace`
`kubecfg env set staging --namespace=staging-namespace`

The default environment will use the namespace of the default context.

This commit will also update commands that take the <env> arg such as
`apply` to make use of the env namespace, if specified.

Fixes #148

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	flagEnvName = "name"
-	flagEnvURI  = "uri"
+	flagEnvName      = "name"
+	flagEnvURI       = "uri"
+	flagEnvNamespace = "namespace"
 )
 
 func init() {
@@ -42,11 +43,15 @@ func init() {
 	// TODO: We need to make this default to checking the `kubeconfig` file.
 	envAddCmd.PersistentFlags().String(flagAPISpec, "version:v1.7.0",
 		"Manually specify API version from OpenAPI schema, cluster, or Kubernetes version")
+	envAddCmd.PersistentFlags().String(flagEnvNamespace, "",
+		"Specify namespace that the environment cluster should use")
 
 	envSetCmd.PersistentFlags().String(flagEnvName, "",
 		"Specify name to rename environment to. Name must not already exist")
 	envSetCmd.PersistentFlags().String(flagEnvURI, "",
 		"Specify URI to point environment cluster to a new location")
+	envSetCmd.PersistentFlags().String(flagEnvNamespace, "",
+		"Specify namespace that the environment cluster should use")
 }
 
 var envCmd = &cobra.Command{
@@ -97,6 +102,11 @@ var envAddCmd = &cobra.Command{
 		envName := args[0]
 		envURI := args[1]
 
+		envNamespace, err := flags.GetString(flagEnvNamespace)
+		if err != nil {
+			return err
+		}
+
 		appDir, err := os.Getwd()
 		if err != nil {
 			return err
@@ -113,7 +123,7 @@ var envAddCmd = &cobra.Command{
 			return err
 		}
 
-		c, err := kubecfg.NewEnvAddCmd(envName, envURI, specFlag, manager)
+		c, err := kubecfg.NewEnvAddCmd(envName, envURI, envNamespace, specFlag, manager)
 		if err != nil {
 			return err
 		}
@@ -146,9 +156,10 @@ environments/
       k8s.libsonnet
       swagger.json
       spec.json      [This will contain the uri of the environment]`,
-	Example: `  # Initialize a new staging environment at us-west. The directory
-  # structure rooted at 'us-west' in the documentation above will be generated.
-  ksonnet env add us-west/staging https://kubecfg-1.us-west.elb.amazonaws.com
+	Example: `  # Initialize a new staging environment at 'us-west'. Using the
+  # namespace 'my-namespace'. The directory structure rooted at 'us-west' in the
+  # documentation above will be generated.
+  ksonnet env add us-west/staging https://kubecfg-1.us-west.elb.amazonaws.com --namespace my-namespace
 
   # Initialize a new staging environment at us-west, using the OpenAPI specification
   # generated in the Kubernetes v1.7.1 build to generate 'ksonnet-lib'.
@@ -225,12 +236,12 @@ var envListCmd = &cobra.Command{
 
 		return c.Run(cmd.OutOrStdout())
 	}, Long: `List all environments in a ksonnet project. This will
-display the name and the URI of each environment within the ksonnet project.`,
+display the name, URI, and namespace of each environment within the ksonnet project.`,
 }
 
 var envSetCmd = &cobra.Command{
-	Use:   "set <env-name> [parameter-flags]",
-	Short: "Set environment fields such as the name, and cluster URI.",
+	Use:   "set <env-name>",
+	Short: "Set environment fields such as the name, cluster URI, and namespace.",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		flags := cmd.Flags()
 		if len(args) != 1 {
@@ -260,7 +271,12 @@ var envSetCmd = &cobra.Command{
 			return err
 		}
 
-		c, err := kubecfg.NewEnvSetCmd(envName, desiredEnvName, desiredEnvURI, manager)
+		desiredEnvNamespace, err := flags.GetString(flagEnvNamespace)
+		if err != nil {
+			return err
+		}
+
+		c, err := kubecfg.NewEnvSetCmd(envName, desiredEnvName, desiredEnvURI, desiredEnvNamespace, manager)
 		if err != nil {
 			return err
 		}
@@ -272,6 +288,9 @@ the name of an environment will also update the directory structure in
 'environments'.`,
 	Example: `  # Updates the URI of the environment 'us-west/staging'.
   ksonnet env set us-west/staging --uri=http://example.com
+
+  # Updates the namespace of the environment 'us-west/staging'.
+  ksonnet env set us-west/staging --namespace=staging
 
   # Updates both the name and the URI of the environment 'us-west/staging'.
   # Updating the name will update the directory structure in 'environments'
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -122,7 +122,7 @@ var RootCmd = &cobra.Command{
 
 // clientConfig.Namespace() is broken in client-go 3.0:
 // namespace in config erroneously overrides explicit --namespace
-func defaultNamespace() (string, error) {
+func namespace() (string, error) {
 	if overrides.Context.Namespace != "" {
 		return overrides.Context.Namespace, nil
 	}
@@ -297,7 +297,8 @@ func parseEnvCmd(cmd *cobra.Command, args []string) (*envSpec, error) {
 //
 // If the environment URI the user is attempting to deploy to is not the current
 // kubeconfig context, we must manually override the client-go --cluster flag
-// to ensure we are deploying to the correct cluster.
+// to ensure we are deploying to the correct cluster. The same logic applies
+// for overwriting the --namespace flag.
 func overrideCluster(envName string) error {
 	cwd, err := os.Getwd()
 	if err != nil {
@@ -335,6 +336,10 @@ func overrideCluster(envName string) error {
 		clusterName := clusterURIs[env.URI]
 		log.Debugf("Overwriting --cluster flag with '%s'", clusterName)
 		overrides.Context.Cluster = clusterName
+		if len(env.Namespace) != 0 {
+			log.Debugf("Overwriting --namespace flag with '%s'", env.Namespace)
+			overrides.Context.Namespace = env.Namespace
+		}
 		return nil
 	}
 
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -40,31 +40,33 @@ const (
 
 // Environment represents all fields of a ksonnet environment
 type Environment struct {
-	Path string
-	Name string
-	URI  string
+	Path      string
+	Name      string
+	URI       string
+	Namespace string
 }
 
 // EnvironmentSpec represents the contents in spec.json.
 type EnvironmentSpec struct {
-	URI string `json:"uri"`
+	URI       string `json:"uri"`
+	Namespace string `json:"namespace"`
 }
 
-func (m *manager) CreateEnvironment(name, uri string, spec ClusterSpec) error {
+func (m *manager) CreateEnvironment(name, uri, namespace string, spec ClusterSpec) error {
 	extensionsLibData, k8sLibData, specData, err := m.generateKsonnetLibData(spec)
 	if err != nil {
 		log.Debugf("Failed to write '%s'", specFilename)
 		return err
 	}
 
-	err = m.createEnvironment(name, uri, extensionsLibData, k8sLibData, specData)
+	err = m.createEnvironment(name, uri, namespace, extensionsLibData, k8sLibData, specData)
 	if err == nil {
-		log.Infof("Environment '%s' pointing to cluster at URI '%s' successfully created", name, uri)
+		log.Infof("Environment '%s' pointing to namespace '%s' and cluster at URI '%s' successfully created", name, namespace, uri)
 	}
 	return err
 }
 
-func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibData, specData []byte) error {
+func (m *manager) createEnvironment(name, uri, namespace string, extensionsLibData, k8sLibData, specData []byte) error {
 	exists, err := m.environmentExists(name)
 	if err != nil {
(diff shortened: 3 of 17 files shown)


metadata/environment.go
```diff
@@ -40,31 +40,33 @@ const (
...
 	extensionsLibData, k8sLibData, specData, err := m.generateKsonnetLibData(spec)
 	if err != nil {
 		log.Debugf("Failed to write '%s'", specFilename)
 		return err
 	}
 
-	err = m.createEnvironment(name, uri, extensionsLibData, k8sLibData, specData)
+	err = m.createEnvironment(name, uri, namespace, extensionsLibData, k8sLibData, specData)
 	if err == nil {
 		log.Infof("Environment '%s' pointing to cluster at URI '%s' successfully created", name, uri)
```
hausdorff: We should consider adding namespace to this message, too. I'd also put namespace first (as I suggest in the `list` comment below), something like:

```
INFO Environment 'default' pointing to namespace 'dev-alex' and cluster at URI '...' successfully created
```

pkg/kubecfg/env.go
```diff
@@ -115,14 +126,16 @@ func (c *EnvListCmd) Run(out io.Writer) error {
...
 }
 
-func NewEnvSetCmd(name, desiredName, desiredURI string, manager metadata.Manager) (*EnvSetCmd, error) {
-	return &EnvSetCmd{name: name, desiredName: desiredName, desiredURI: desiredURI, manager: manager}, nil
+func NewEnvSetCmd(name, desiredName, desiredURI, desiredNamespace string, manager metadata.Manager) (*EnvSetCmd, error) {
+	return &EnvSetCmd{name: name, desiredName: desiredName, desiredURI: desiredURI, desiredNamespace: desiredNamespace,
+		manager: manager}, nil
 }
 
 func (c *EnvSetCmd) Run() error {
```
hausdorff: If I'm not mistaken, we're forgetting to set the `Namespace` member of the `Environment` object in this method. Meaning that any call to `env set --namespace="whatever"` will result in setting the namespace as `"default"`.

metadata/environment.go
```diff
@@ -279,24 +276,38 @@ func (m *manager) SetEnvironment(name string, desired *Environment) error {
...
+		URI = desired.URI
+	} else {
+		URI = env.URI
+	}
+	var namespace string
+	if len(desired.Namespace) != 0 {
+		log.Infof("Setting environment namespace to '%s'", desired.Namespace)
+		namespace = desired.Namespace
+	} else {
+		namespace = env.Name
```
hausdorff: If I'm not mistaken, this will set the namespace to the `Name`. So if I have some env with a name `foo`, and I run something like `env set --namespace=""` we'll set the namespace to `"foo"`.

I'm not sure we want to distinguish between `--namespace=""` and the `--namespace` flag not being present at all. One option is to just have `--namespace=""` result in no changes, and another is to have that "unset" the variable to either `""` or `"default"`.

cmd/env.go
```diff
@@ -144,9 +154,10 @@ environments/
       k8s.libsonnet
       swagger.json
       spec.json      [This will contain the uri of the environment]`,
-	Example: `  # Initialize a new staging environment at us-west. The directory
-  # structure rooted at 'us-west' in the documentation above will be generated.
-  ksonnet env add us-west/staging https://kubecfg-1.us-west.elb.amazonaws.com
+	Example: `  # Initialize a new staging environment at 'us-west'. Using the
+  # namespace 'my-namespace'. The directory structure rooted at 'us-west' in the
+  # documentation above will be generated.
+  ksonnet env add us-west/staging https://kubecfg-1.us-west.elb.amazonaws.com --namespace my-namespace
```
hausdorff: So. I am thinking we also want to enable a flag `--context`, which has semantics much like `init`. For example, the following example would automatically init namespace and URI from some context.

```
ks env add us-west/staging --context your-context
```

jessicayuen: Yea, this would be a useful feature. I'll probably open an issue to add this as a feature, or add it as a task item to #165. Thoughts?

pkg/kubecfg/env.go
```diff
@@ -98,10 +100,19 @@ func (c *EnvListCmd) Run(out io.Writer) error {
...
+			maxUriLen = uriLen
+		}
+	}
+
 	lines := []string{}
 	for _, env := range envs {
 		nameSpacing := strings.Repeat(" ", maxNameLen-len(env.Name)+1)
-		lines = append(lines, env.Name+nameSpacing+env.URI+"\n")
+		uriSpacing := strings.Repeat(" ", maxUriLen-maxNameLen-len(env.URI))
+		lines = append(lines, env.Name+nameSpacing+env.URI+uriSpacing+env.Namespace+"\n")
```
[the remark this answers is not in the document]
jessicayuen: Ooo both good points, SGTM!

pkg/kubecfg/env.go
```diff
@@ -85,23 +92,37 @@ func (c *EnvListCmd) Run(out io.Writer) error {
...
 	for _, env := range envs {
 		if l := len(env.Name); l > maxNameLen {
 			maxNameLen = l
 		}
 	}
 
+	maxNamespaceLen := len(namespaceHeader) + maxNameLen + 1
+	for _, env := range envs {
+		if l := len(env.Namespace) + maxNameLen + 1; l > maxNamespaceLen {
+			maxNamespaceLen = l + maxNameLen + 1
```
hausdorff: This doesn't seem right? It looks like we're adding `maxNameLen + 1` twice.

anguslees:
Does this just set the default (fallback) namespace? (ie: what `-n` does currently)

In particular: what happens when objects declare an explicit namespace?  how will this feature behave when the objects span multiple namespaces, and/or includes non-namespaced objects (CRD declarations, ClusterRoles, etc)?  I see the code just sets the equivalent of existing `-n`, but various symbols are renamed `DefaultNamespace` -> `Namespace` and other changes that make the namespace more prominent - so I guess I'm wondering what the future intended semantics are here.

(I'd kind of like to see the reverse version, where we remove the legacy kubectl idea of an implicit "ambient" namespace, and instead make it always explicit via a convention around regular jsonnet parameters)

seh:
[I brought up a related question](https://github.com/ksonnet/kubecfg/issues/148#issuecomment-331666984) in #148.

I agree with @anguslees that the default/overridden namespace treatment in _kubectl_ is contentious, at times confusing or annoying. If you try to deal with manifests that already have a "metadata.namespace" field, you can leave off _kubectl_'s `--namespace` flag or match the values of the "metadata.namespace" fields, which feels like a cruel game of needing to restate a predetermined answer.

In my project, I mandated that we maintain most of our Kubernetes manifest files with no "metadata.namespace" field, and that we'd use the `--namespace` flag with _kubectl apply_ (or _kubectl create_) to project those manifests into one or more appropriate namespaces. However, that approach limits how you can use _kubectl apply --prune_, because you're more likely to want to throw files from multiple namespaces into the set considered for pruning. I'm now experimenting with going back the other way, with every manifest having a populated "metadata.namespace" field, and duplicating manifests that we used to share into a separate set for each namespace, backing away from relying on the `--namespace` flag.

That's still a separate matter from the namespace nominated in the KUBECONFIG file as "contexts.context.namespace." From both #148 and the changes considered here, we're considering adding something more like _kubectl_'s `--namespace` flag to override the namespace designated by the current context.

Could we get away with @anguslees's "reverse version," where there's no built-in way to confine the target namespace?

hausdorff:
@anguslees @seh The core issue right now is that an environment points at a specific cluster URI, but the namespace is populated by client-go. So, if you run `apply us-west/dev` and you happen to have `namespace` set in your `$KUBECONFIG`, the cluster points at `us-west/dev` but the namespace is pulled from current-context. This is quite likely to be not what you want.

In order for these commands to make sense, I think we to either:

1. Always override the `namespace` to be `""` if it's populated from `$KUBECONFIG`, or
1. Allow users to specify their own `namespace` as part of the environment specification.

In the case of this PR, we're proposing to auto-populate `namespace` with whatever's in `$KUBECONFIG`'s current-context. I could see us defaulting to `""`, but to not have any ability at all to specify namespace at the ksonnet app level, I think, is going to be a point of friction.

seh:
If you're not honoring the API server URL from _KUBECONFIG_, then you should not honor a namespace specified there either. That leans toward knocking it out with either "", or some weird sentinel like "ksonnet-nonexistent."

For cases where you're managing several namespaces in a cluster, specifying a singular "target" namespace doesn't make any sense. For cases where you're managing a single namespace, then it does make sense, and could even be necessary if the materialized Kubernetes manifests lack a "metadata.namespace" field. However, you _could_ instead mandate that that field be populated in order to have this tool use such manifests.

hausdorff:
I completely agree that there are applications where we don't want to have namespace set, and this proposal does support that.

But, another goal is to work out of the box with existing k8s apps, so I'm uncomfortable not supporting namespace as part of an environment spec, since a nontrivial set of people would have to change a lot of code in order to get the desired effect. Besides, we also allow the `--namespace` flag, so it's not like we don't allow people to specify this anyway.

seh:
I didn't mean to suggest that you eliminate a target namespace from the environment specification. I wanted to make it clear that an environment should completed occlude using values out of a _KUBECONFIG_ file.

If the environment can designate a target namespace, but not require that one be set, that would satisfy my concern. I'd like it to behave such that if I set no namespace in my environment, and I attempt (accidentally) to operate on an object whose manifests lacks a populated "metadata.namespace" field, the operation would fail.

anguslees:
I think we're all in firm agreement on the big picture: if we are going to supply an ambient namespace, then it should be from environment not ~/.kube/config.  So in that sense, lgtm.

@jessicayuen: My understanding is that the current code *does* use the namespace from the current active KUBECONFIG context (if there is no more-explicit `-n` cli arg), and I thought [this test](https://github.com/ksonnet/kubecfg/blob/master/integration/kubeflags_test.go#L91) confirmed that behaviour.  Do you see otherwise? (is the test broken/insufficient?)

@seh: The kubecfg code has to _explicitly_ handle missing metadata.namespace and substitute the default namespace (for namespaced resources).  So it would be very easy, and robust, to change this behaviour to (eg) abort with an error instead if there was an object with blank/missing metadata.namespace.  I agree with @hausdorff that this would be surprising/upsetting to people who relied on this legacy implicit namespace behaviour of `kubectl`, but it's certainly something we could do.  (I think we have both groups internally at Bitnami fwiw - I know of one repo that relies on late-bound namespace for testing vs prod deployments, and another that has a CI test that rejects PRs with missing metadata.namespace)

seh:
Thank you for clarifying, @anguslees. I would like it if one could express in an environment either of two preferences:

• When a namespaced kind's manfiest's "metadata.namespace" field is absent, use this namespace.
• When a namespaced kind's manifest's "metadata.namespace" field is absent, fail.

We could implement that by treating a present, nonempty namespace in the environment as choosing the first preference, and the lack of a namespace in the environment as choosing the latter preference. Calling the environment field something like "default namespace"—though confusing due to the ever-present "default" Kubernetes namespace—would make it clear that the namespace is only used to avoid a failure that would otherwise occur.

jessicayuen:
@anguslees : I meant that the current code uses the current active context if there is no explicit (or implicit in the case of this PR) -n flag set. It is when the -n flag is set that `defaultNamespace()` makes less sense. It is even more so the case with this change, where the -n flag is set implicitly if there is a namespace specified in the environment.

cmd/env.go
```diff
@@ -40,11 +41,15 @@ func init() {
 	// TODO: We need to make this default to checking the `kubeconfig` file.
 	envAddCmd.PersistentFlags().String(flagAPISpec, "version:v1.7.0",
 		"Manually specify API version from OpenAPI schema, cluster, or Kubernetes version")
+	envAddCmd.PersistentFlags().String(flagEnvNamespace, "",
+		"Specify namespace that the environment cluster should use")
```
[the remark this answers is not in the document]
hausdorff: I have an open issue to think through this.

metadata/environment.go
```diff
@@ -40,31 +40,33 @@ const (
...
+		log.Infof("Environment '%s' pointing to namespace '%s' and cluster at URI '%s' successfully created", name, namespace, uri)
 	}
 	return err
 }
 
-func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibData, specData []byte) error {
+func (m *manager) createEnvironment(name, uri, namespace string, extensionsLibData, k8sLibData, specData []byte) error {
 	exists, err := m.environmentExists(name)
 	if err != nil {
 		log.Debug("Failed to check whether environment exists")
```
hausdorff: Below, we should probably also change the logging message to include the namespace. From:

```go
log.Infof("Creating environment '%s' with uri '%s'", name, uri)
```

To something like:

```go
```
log.Infof("Creating environment '%s' with namespace '%s', pointing at cluster located at '%s'", name, namespace, uri)
```
```

cmd/root.go
```diff
@@ -324,6 +325,10 @@ func overrideCluster(envName string) error {
 		clusterName := clusterURIs[env.URI]
 		log.Debugf("Overwriting --cluster flag with '%s'", clusterName)
 		overrides.Context.Cluster = clusterName
+		if len(env.Namespace) != 0 {
+			log.Debugf("Overwriting --namespace flag with '%s'", env.Namespace)
+			overrides.Context.Namespace = env.Namespace
+		}
```
hausdorff: Sorry, I forget what we decided here. When a user runs something like `env set foo --namespace=""`, we have the following options:

* Unset the namespace.
* Do nothing.
* Set namespace to `"default"`.

I'm fine pushing this to an issue and punting for now; I just want to make sure we make this decision purposefully. So why don't we do that?

That said, my inclination is to set it to `"default"`. We also might also consider having an `env unset` command.

metadata/environment.go
```diff
@@ -279,24 +276,38 @@ func (m *manager) SetEnvironment(name string, desired *Environment) error {
...
+	} else {
+		URI = env.URI
+	}
+	var namespace string
+	if len(desired.Namespace) != 0 {
+		log.Infof("Setting environment namespace to '%s'", desired.Namespace)
+		namespace = desired.Namespace
+	} else {
+		namespace = env.Namespace
+	}
```
hausdorff: I notice that we do the length == 0 check twice. I think if we do settle on checking it, it should be here, so it's consistent across all call sites, rather than in the command?

pkg/kubecfg/env.go
```diff
@@ -86,22 +93,37 @@ func (c *EnvListCmd) Run(out io.Writer) error {
...
 			maxNameLen = l
 		}
 	}
 
+	maxNamespaceLen := len(namespaceHeader) + maxNameLen + 1
+	for _, env := range envs {
+		if l := len(env.Namespace) + maxNameLen + 1; l > maxNamespaceLen {
+			maxNamespaceLen = l
+		}
+	}
```
hausdorff: Can we make an issue to write a helper routine to space-pad columns here? We're doing it in a bunch of places and we've gotten it wrong at least once, so I think now's the time to pull the trigger.

jessicayuen: opened #171

metadata/environment.go
jessicayuen: wrong line, but fixed. Will push up changes soon.

cmd/root.go
jessicayuen: I'll put this into an issue (or if I have time later today fix it), but I feel that `"default"` might not be the right answer -- what if the cluster actually has a namespace called default?  I think env unset is better, or allow them to do `env set foo --namespace=""`.

metadata/environment.go
jessicayuen: So, I think it's worth checking in both places because in this location, it's ensuring we're not setting empty namespaces into the `spec.json` file.  

The other place is where we're actually overriding the cluster flag with some environment's namespace. If the env didn't have a namespace specified, and the user put `-n` flag through the command line, we'd be overwriting it with "". I'm not sure that'd be desirable.

Thoughts?

hausdorff: Ok, that seems fine. We should mention that in the issue below, where we deal with empty namespaces.

cmd/root.go
jessicayuen: My mistake -- did not realize `default` exists as a namespace. Opened issue: #172

jessicayuen:
Closing -- moved to the [ksonnet repository](https://github.com/ksonnet/ksonnet/pull/17).

PR #164, opened 2017-09-28
Construct base components Obj

jessicayuen:
This implements the `ksonnet.importComponents()` piece specified by the design [here](https://docs.google.com/document/d/1yU6Zi3t1D-sSLnMEE6aQn2NS41tLqiYgzEoZnQvQEbY).

In order to support environment heirarchy, we need to be able to
reference all components. This commit will implement component imports
as k-v pairs ex: foo: "import/foo.jsonnet" in a Jsonnet object. This
jsonnet object will then be assigned using ExtCode so that it can be
referenced by a base.libsonnet file which environments can build /
override on.

See the design doc above for more information.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -58,6 +59,8 @@ const (
 	// environment or the -f flag.
 	flagFile      = "file"
 	flagFileShort = "f"
+
+	componentsExtCodeKey = "__ksonnet/components"
 )
 
 var clientConfig clientcmd.ClientConfig
@@ -369,8 +372,9 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
 			if err != nil {
 				return nil, err
 			}
+			baseObjExtCode := fmt.Sprintf("%s=%s", componentsExtCodeKey, constructBaseObj(fileNames))
+			expander.ExtCodes = append([]string{baseObjExtCode})
 		}
-
 	}
 
 	//
@@ -378,4 +382,26 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
 	//
 
 	return expander.Expand(fileNames)
+}
+
+// constructBaseObj constructs the base Jsonnet object that represents k-v
+// pairs of component name -> component imports. For example,
+//
+//   {
+//      foo: import "components/foo.jsonnet"
+//   }
+func constructBaseObj(paths []string) string {
+	var obj bytes.Buffer
+	obj.WriteString("{\n")
+	for _, p := range paths {
+		ext := path.Ext(p)
+		if path.Ext(p) != ".jsonnet" {
+			continue
+		}
+
+		name := strings.TrimSuffix(path.Base(p), ext)
+		fmt.Fprintf(&obj, "  %s: import \"%s\",\n", name, p)
+	}
+	obj.WriteString("}\n")
+	return obj.String()
 }
--- a/template/expander.go
+++ b/template/expander.go
@@ -20,6 +20,7 @@ type Expander struct {
 	ExtVarFiles []string
 	TlaVars     []string
 	TlaVarFiles []string
+	ExtCodes    []string
 
 	Resolver   string
 	FailAction string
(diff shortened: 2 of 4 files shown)


cmd/root.go
```diff
@@ -170,7 +173,7 @@ func (f *logFormatter) Format(e *log.Entry) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func newExpander(cmd *cobra.Command) (*template.Expander, error) {
+func newExpander(cmd *cobra.Command, cwd metadata.AbsPath) (*template.Expander, error) {
```
hausdorff: I think we don't need the `cwd` argument anymore? Here and in `expandEnvCmdObjs`.

cmd/root.go
```diff
@@ -379,3 +382,26 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
...
+	lines := []string{}
+	lines = append(lines, "{\n")
+	for _, p := range paths {
+		ext := path.Ext(p)
+		if path.Ext(p) != jsonnetExt {
+			continue
+		}
+
+		name := strings.TrimSuffix(path.Base(p), ext)
+		lines = append(lines, "  "+name+": import \""+p+"\",\n")
```
hausdorff: Might be clearer with something like:

```
fmt.Sprintf("  %s: import \"%s\",\n", name, p)
```

cmd/root.go
```diff
@@ -379,3 +382,26 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
 
 	return expander.Expand(fileNames)
 }
+
+// constructBaseObj constructs the base Jsonnet object that represent k-v
```
seh: - [ ] s/represent/represents/

cmd/root.go
```diff
@@ -379,3 +382,26 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
...
+// pairs of component name -> component imports.
+// For example,
+// {
+//    foo: import "components/foo.jsonnet"
+// }
+func constructBaseObj(paths []string) string {
+	lines := []string{}
```
seh: - [ ] Consider using a `bytes.Buffer` to accumulate the lines here, then return them all via [`(*bytes.Buffer).String`](https://golang.org/pkg/bytes/#Buffer.String).
```go
var obj bytes.Buffer // or bytes.NewBuffer(make([]byte, 512))
obj.WriteString("{\n")
// ...
obj.WriteByte('}')
return obj.String()
```

jessicayuen: Thanks for the review, @seh! Fixed.

cmd/root.go
```diff
@@ -379,3 +382,25 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
 
 	return expander.Expand(fileNames)
 }
+
+// constructBaseObj constructs the base Jsonnet object that represents k-v
+// pairs of component name -> component imports.
+// For example,
+// {
+//    foo: import "components/foo.jsonnet"
+// }
```
hausdorff: Tiny nit, we tend to indent code blocks embedded in comments. Also, `For example,` looks like it fits on the previous line?

```
// constructBaseObj constructs the base Jsonnet object that represents k-v
// pairs of component name -> component imports. For example,
//
//   {
//      foo: import "components/foo.jsonnet"
//   }
```

template/expander.go
```diff
@@ -21,6 +25,8 @@ type Expander struct {
 	TlaVars     []string
 	TlaVarFiles []string
 
+	BaseObj string
```
hausdorff: We might consider having a member called `ExtCode` instead of `BaseObj`. Maybe this means making a new type, like `[]ExtCode`.

Personally, I'm a little bit wary of having `BaseObj` baked directly into `template.Expander`, since it was meant to be a reasonably faithful expression of the Jsonnet VM API, until we can replace the cgo implementation with pure Go implementation.

cmd/root.go
```diff
@@ -379,3 +382,25 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
...
+	var obj bytes.Buffer
+	obj.WriteString("{\n")
+	for _, p := range paths {
+		ext := path.Ext(p)
+		if path.Ext(p) != jsonnetExt {
+			continue
+		}
+
+		name := strings.TrimSuffix(path.Base(p), ext)
+		obj.WriteString(fmt.Sprintf("  %s: import \"%s\",\n", name, p))
```
seh: Here, you can use [`fmt.Fprintf`](https://golang.org/pkg/fmt/#Fprintf) and supply a pointer to "obj" as your `io.Writer` instance, saving you the intermediate string created by `fmt.Sprintf`.

Issue #165, opened 2017-09-28
Implement "complex" environments


jessicayuen:
"complex" environments, are based off this [design](https://docs.google.com/document/d/1yU6Zi3t1D-sSLnMEE6aQn2NS41tLqiYgzEoZnQvQEbY).

- - [ ] **PR #164** Add intrinsic for `ksonnet.importComponents()`
- - [ ] **PR #167** Update directory structure
    ```
    environments/
      base.libsonnet
      us-east/
         dev/
           dev.jsonnet
           spec.json
           .metadata/
              k8s.libsonnet
              k.libsonnet
      ```
- - [ ] **PR #167** Generate `base.libsonnet`
    ```
    local components = ksonnet.importComponents();
    components + {
      // Insert user-specified overrides here.
    }
    ```
- - [ ] **PR #167** Generate `.libsonnet` file on `env add`
   ```
   local base = import "base.libsonnet";
   local k = import "k.libsonnet";
   local dep = k.apps.v1beta1.deployment;

   base + {
      // Insert user-specified overrides here
   }
   ```
- - [ ] **PR #167** Commands that take env need to apply `<env>.libsonnet` file
- - [ ] **PR #169** Support multi-env diff #163 

- - [ ] Automatically init namespace and URI from some context. `ks env add us-west/staging --context your-context`

More to be added to support params

jessicayuen:
This issue was moved to ksonnet/ksonnet#6

PR #167, opened 2017-10-02
Environment base.libsonnet and overrides

jessicayuen:
Implements part of #165. This PR does a few things:

1. Hide environment metadata details that we do not want users modifying in a `.metadata` folder (`k.libsonnet`, `k8s.libsonnet`, `swagger.json`)

2. Generate base.libsonnet:
base.libsonnet is a generated file that exists at the root of the
environments directory. This file is generated for all ksonnet projects.
The main goal of this file is to import all components in the components
directory, so that environments are able to easily extend / override any
one of these components in a modular structure.

3. Generate the override file:
Generate an "<env-name>.jsonnet" file when an environment is created.
The goal of this file is to allow users to extend on base.libsonnet on a
per-environment basis to allow for custom overrides, such as replica
count.
For example, the default environment would have the following tree
structure:
```
├── environments
│   ├── base.libsonnet
│   └── default
│       ├── .metadata
│       │   ├── k.libsonnet
│       │   ├── k8s.libsonnet
│       │   └── swagger.json
│       ├── default.jsonnet
│       └── spec.json
```

4. Expand the override file:
Commands that take `env` as a param currently expand all files in the
`components` directory. This is no longer necessary with the
introduction of `base.libsonnet` and the per-environment override
`<env>.jsonnet` file. This change will simply expand the single
`<env>.jsonnet` file (which will implicitly expand all component files).
The case of running `ksonnet apply default`, is equivalent to running
`ksonnet apply -f environments/default/default.jsonnet`.

### DIFF
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -30,7 +30,8 @@ import (
 )
 
 const (
-	defaultEnvName = "default"
+	defaultEnvName  = "default"
+	metadataDirName = ".metadata"
 
 	schemaFilename        = "swagger.json"
 	extensionsLibFilename = "k.libsonnet"
@@ -87,11 +88,17 @@ func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibD
 		return err
 	}
 
+	metadataPath := appendToAbsPath(envPath, metadataDirName)
+	err = m.appFS.MkdirAll(string(metadataPath), defaultFolderPermissions)
+	if err != nil {
+		return err
+	}
+
 	log.Infof("Generating environment metadata at path '%s'", envPath)
 
 	// Generate the schema file.
 	log.Debugf("Generating '%s', length: %d", schemaFilename, len(specData))
-	schemaPath := appendToAbsPath(envPath, schemaFilename)
+	schemaPath := appendToAbsPath(metadataPath, schemaFilename)
 	err = afero.WriteFile(m.appFS, string(schemaPath), specData, defaultFilePermissions)
 	if err != nil {
 		log.Debugf("Failed to write '%s'", schemaFilename)
@@ -99,7 +106,7 @@ func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibD
 	}
 
 	log.Debugf("Generating '%s', length: %d", k8sLibFilename, len(k8sLibData))
-	k8sLibPath := appendToAbsPath(envPath, k8sLibFilename)
+	k8sLibPath := appendToAbsPath(metadataPath, k8sLibFilename)
 	err = afero.WriteFile(m.appFS, string(k8sLibPath), k8sLibData, defaultFilePermissions)
 	if err != nil {
 		log.Debugf("Failed to write '%s'", k8sLibFilename)
@@ -107,7 +114,7 @@ func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibD
 	}
 
 	log.Debugf("Generating '%s', length: %d", extensionsLibFilename, len(extensionsLibData))
-	extensionsLibPath := appendToAbsPath(envPath, extensionsLibFilename)
+	extensionsLibPath := appendToAbsPath(metadataPath, extensionsLibFilename)
 	err = afero.WriteFile(m.appFS, string(extensionsLibPath), extensionsLibData, defaultFilePermissions)
 	if err != nil {
 		log.Debugf("Failed to write '%s'", extensionsLibFilename)
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -123,7 +123,7 @@ func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibD
 		return err
 	}
 
-	// Generate the base.libsonnet override file
+	// Generate the environment .jsonnet file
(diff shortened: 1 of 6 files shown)


metadata/environment.go
```diff
@@ -30,12 +32,15 @@ import (
...
-	defaultEnvName = "default"
+	defaultEnvName  = "default"
+	metadataDirName = ".metadata"
 
 	schemaFilename        = "swagger.json"
 	extensionsLibFilename = "k.libsonnet"
 	k8sLibFilename        = "k8s.libsonnet"
 	specFilename          = "spec.json"
+
+	libsonnetExt = ".libsonnet"
```
hausdorff: Micro nit, as I've mentioned in previous reviews, I think we probably don't need this, since it's not going to change, we use it once, and other parts of the code just use the string `".libsonnet"`.

metadata/environment.go
```diff
@@ -327,6 +349,17 @@ func (m *manager) generateKsonnetLibData(spec ClusterSpec) ([]byte, []byte, []by
 	return extensionsLibData, k8sLibData, text, err
 }
 
+func (m *manager) generateOverrideData() []byte {
+	var buf bytes.Buffer
+	buf.WriteString(fmt.Sprintf("local base = import \"%s\";\n", m.baseLibsonentPath))
+	buf.WriteString(fmt.Sprintf("local k = import \"%s\";\n", path.Join(metadataDirName, extensionsLibFilename)))
+	buf.WriteString("local dep = k.apps.v1beta1.deployment;\n\n")
```
hausdorff: I'm not sure we need this line?

metadata/manager.go
```diff
@@ -38,6 +38,8 @@ const (
 	componentsDir   = "components"
 	environmentsDir = "environments"
 	vendorDir       = "vendor"
+
+	baseLibsonnetFile = "base.libsonnet"
```
hausdorff: Typo?

metadata/environment.go
```diff
@@ -30,7 +30,8 @@ import (
 )
 
 const (
-	defaultEnvName = "default"
+	defaultEnvName  = "default"
+	metadataDirName = ".metadata"
```
hausdorff: This change also needs to `Manager.LibPaths` to return the correct path to these builtin libraries. Otherwise commands like `show`, `apply`, _etc_., will fail.

metadata/manager.go
jessicayuen: I can't see it! where?

metadata/environment.go
jessicayuen: I fixed this, but also added a test for the LibPaths().

metadata/manager.go
hausdorff: Update: this was resolved. I clicked the wrong line.

metadata/environment.go
```diff
@@ -334,6 +347,16 @@ func (m *manager) generateKsonnetLibData(spec ClusterSpec) ([]byte, []byte, []by
 	return extensionsLibData, k8sLibData, text, err
 }
 
+func (m *manager) generateOverrideData() []byte {
+	var buf bytes.Buffer
+	buf.WriteString(fmt.Sprintf("local base = import \"%s\";\n", m.baseLibsonnetPath))
+	buf.WriteString(fmt.Sprintf("local k = import \"%s\";\n\n", path.Join(metadataDirName, extensionsLibFilename)))
```
hausdorff: I think the `-J` flags allow you to just make this `import "k.libsonnet"`, rather than `import ".metadata/k.libsonnet"`?

metadata/environment.go
```diff
@@ -334,6 +347,16 @@ func (m *manager) generateKsonnetLibData(spec ClusterSpec) ([]byte, []byte, []by
...
+	buf.WriteString("base + {\n")
+	buf.WriteString("  // Insert user-specified overrides here.\n")
```
hausdorff: Might add an example:

```
  // Insert user-specific overrides here. For example if a component is named "nginx-deployment", you might have something like:
  //   "nginx-deployment"+: k.deployment.mixin.metadata.labels({foo: "bar"})
```

metadata/environment.go
```diff
@@ -121,6 +123,17 @@ func (m *manager) createEnvironment(name, uri string, extensionsLibData, k8sLibD
 		return err
 	}
 
+	// Generate the base.libsonnet override file
```
hausdorff: Really this is generating the environment file. Maybe something like `Generate the environment .jsonnet file`.

metadata/manager.go
```diff
@@ -190,11 +196,20 @@ func (m *manager) createAppDirTree() error {
...
 			return err
 		}
 	}
 
-	return nil
+	return afero.WriteFile(m.appFS, string(m.baseLibsonnetPath), genBaseLibsonnetContent(), defaultFilePermissions)
+}
+
+func genBaseLibsonnetContent() []byte {
+	// TODO replace with constant ref once #164 is merged
```
hausdorff: Is this TODO still relevant?

jessicayuen:
Closing -- moved PR to new [ksonnet repository](https://github.com/ksonnet/ksonnet/pull/18).

PR #176, opened 2017-11-17
Revert ksonnet/ksonnet-specific changes since last kubecfg release

anguslees:
This change is a big rewind and selective cherrypick of just the kubecfg-specific changes since the last kubecfg release.  "ksonnet" has now been forked into its own project, and this completes the kubecfg-side of the fork process.

Selectively picking commits is inherently error prone, we might have some further cleanup to do.  I am super-grateful for all the excellent commit descriptions that made something like this achievable.

/cc @hausdorff @jessicayuen


anguslees:
Note there is one original commit in this stack: I switched Trim -> TrimSuffix in the server minor version code.

anguslees:
Oh and if anyone reading this is missing any of the code/features removed in this PR, that probably means they should switch to the [ksonnet/ksonnet](https://github.com/ksonnet/ksonnet) codebase.

PR #180, opened 2017-12-15
[jjo] diff: handle empty config values

### LINKED ISSUE #179: diff should skip empty config values, as api won't return them
Example from creating a pod from scratch (note empty `env`):
```
$ cat nginx.pod.yaml 
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: nginx
  name: nginx
spec:
  containers:
  - image: nginx
    env: []
    name: nginx
  restartPolicy: Never

$ kubecfg update -v nginx.pod.yaml
INFO  Updating pods nginx  
INFO  Creating non-existent pods nginx

$ kubecfg diff --diff-strategy subset nginx.pod.yaml
---
- live pods nginx
+ config pods nginx
 {
   "apiVersion": "v1",
   "kind": "Pod",
   "metadata": {
     "labels": {
       "run": "nginx"
     },
     "name": "nginx"
   },
   "spec": {
     "containers": [
       {
         "image": "nginx",
         "imagePullPolicy": "IfNotPresent",
         "name": "nginx",  
         "resources": {
         }
+        "env": [
+        ]
       }
     ],
     "dnsPolicy": "ClusterFirst",
     "restartPolicy": "Never"
   }
 }
ERROR Differences found.  
```

jjo:
Fixes #179.

### DIFF
--- a/pkg/kubecfg/diff.go
+++ b/pkg/kubecfg/diff.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"reflect"
 	"sort"
 
 	isatty "github.com/mattn/go-isatty"
@@ -102,6 +103,26 @@ func (c DiffCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer) err
 	return nil
 }
 
+// isEmptyValue is taken from the encoding/json package in the
+// standard library.
+func isEmptyValue(v reflect.Value) bool {
+	switch v.Kind() {
+	case reflect.Array, reflect.Map, reflect.Slice, reflect.String:
+		return v.Len() == 0
+	case reflect.Bool:
+		return !v.Bool()
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return v.Int() == 0
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		return v.Uint() == 0
+	case reflect.Float32, reflect.Float64:
+		return v.Float() == 0
+	case reflect.Interface, reflect.Ptr:
+		return v.IsNil()
+	}
+	return false
+}
+
 func removeFields(config, live interface{}) interface{} {
 	switch c := config.(type) {
 	case map[string]interface{}:
@@ -118,6 +139,9 @@ func removeMapFields(config, live map[string]interface{}) map[string]interface{}
 	for k, v1 := range config {
 		v2, ok := live[k]
 		if !ok {
+			if isEmptyValue(reflect.ValueOf(v1)) {
+				result[k] = v1
+			}
 			continue
 		}
 		result[k] = removeFields(v1, v2)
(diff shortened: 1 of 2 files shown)


pkg/kubecfg/diff.go
```diff
@@ -102,6 +103,26 @@ func (c DiffCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer) err
 	return nil
 }
 
+// isEmptyValue is taken from the encoding/json package in the
```
mmikulicic: please mention https://github.com/golang/go/issues/7501

pkg/kubecfg/diff.go
```diff
@@ -118,6 +139,9 @@ func removeMapFields(config, live map[string]interface{}) map[string]interface{}
 	for k, v1 := range config {
 		v2, ok := live[k]
 		if !ok {
+			if isEmptyValue(reflect.ValueOf(v1)) {
```
mmikulicic: perhaps a comment wouldn't hurt here (so one doesn't have to git blame through the full future chain of commits until one hits this PR with the relevant context)

pkg/kubecfg/diff.go
jjo: +1, done.

pkg/kubecfg/diff.go
jjo: +1, done.

pkg/kubecfg/diff.go
```diff
@@ -102,6 +103,28 @@ func (c DiffCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer) err
 	return nil
 }
 
+// isEmptyValue is taken from the encoding/json package in the
+// standard library.
+// See also feature request for golang reflect pkg at
+// https://github.com/golang/go/issues/7501
+func isEmptyValue(v reflect.Value) bool {
+	switch v.Kind() {
```
anguslees: Fwiw, this only has to deal with the types that `encoding/json` produces, so a simple typeswitch will be faster:

```golang
func isEmptyValue(i interface{}) bool {
        switch v := i.(type) {
        case []interface{}:
                return len(v) == 0
        case map[string]interface{}:
                return len(v) == 0
        case bool:
                return !v
        case float64:
                return v == 0
        case string:
                return v == ""
        case nil:
                return true
        default:
                panic("Found unexpected type %T in json unmarshal (value=%v)", i, i)
        } 
}
```

jjo: @anguslees: nopes, that won't support nested `nil`s - I've augmented
`diff_test.go` a bit to exercise that also.

FWIW using above simpler implementation would expectedly fail with:
```
# github.com/ksonnet/kubecfg/pkg/kubecfg
pkg/kubecfg/diff.go:142:8: too many arguments to panic: panic("Found unexpected type %T in json unmarshal (value=%v)", i, i)
Makefile:35: recipe for target 'kubecfg' failed
```

jjo: @anguslees nvm on above reply (early morning low caffeine disclaimer :P),
re-implemented it using your suggestions above (+couple nits).

Issue #19, opened 2017-06-01
Include windows executable in releases


anguslees:
Build/test/release successfully on windows.

PR #184, opened 2018-02-08
Use go-jsonnet instead of jsonnet_go

### LINKED ISSUE #183: Use go-jsonnet
As discussed in https://github.com/ksonnet/kubecfg/issues/182#issuecomment-364195847, there's a desire to ditch the CGO-based `jsonnet_go` with `go-jsonnet`, which should now be on par with the native C implementation.

petr-k:
Closes #183 

Separated into two commits for easier review.


anguslees:
Love it. There are some release build details that we can simplify now that we don't need cgo, but I'll do those in followup PRs.

petr-k:
Plus, now that CGO's gone, we can now easily cross-compile for Windows, right? That would solve #19.

anguslees:
> Plus, now that CGO's gone, we can now easily cross-compile for Windows, right? That would solve #19.

With some effort, we could have cross-compiled using cgo too.  I avoided cross-compiling previously because I wanted to actually execute the binary as part of testing.  Most of that was because I had encountered frequent toolchain failures on osx however - now that we're regular-go, we can probably be more confident that cross-built binaries will just work...

PR #187, opened 2018-02-11
Support URLs in imports and search paths

### LINKED ISSUE #182: Wishlist: support http/s jsonnet search path
It would be great if the jsonnet lib search path supported locating remote resources via HTTP. My use case involves keeping some jsonnet libraries in separate repositories in GitLab (where raw files can be accessed via HTTP).

While it is possible to download the libraries to disk and adjusting the search path, it requires wrapping `kubecfg` with custom code. It would be nice if `kubecfg` supported this directly.

Thoughts:
* if there are any concerns about this breaking existing jsonnet files out there, it might be turned on only via a cmdline flag
* could be implemented by overriding ksonnet VM's `ImportCallback`
  - the callback would first delegate to the original callback, searching the filesystem
  - then it could try `GET`ing a jsonnet document from the webserver; url would be constructed by joining the search path url (beginning with `http://`) and the import path. If a non-success code is returned, nothing happens (except logging this), otherwise the response contents is converted to string and imported as jsonnet.

@anguslees Does this make sense to you? Would you accept a PR implementing this functionality?

petr-k:
Implements an importer that supports (in addition to standard filesystem imports)
  - URLs in import statements
  - URLs in library search paths

A real-world example:
  - you have https://raw.githubusercontent.com/ksonnet/ksonnet-lib/master in your search URLs
  - you evaluate a local file which calls `import "ksonnet.beta.2/k.libsonnet"`
  - if the `ksonnet.beta.2/k.libsonnet` is not located in the current workdir, an attempt
    will be made to follow the search path, i.e. to download
    https://raw.githubusercontent.com/ksonnet/ksonnet-lib/master/ksonnet.beta.2/k.libsonnet
  - since the downloaded `k.libsonnet` file turn in calls `import "k8s.libsonnet"`, the import
    will be resolved as https://raw.githubusercontent.com/ksonnet/ksonnet-lib/master/ksonnet.beta.2/k8s.libsonnet
	and downloaded from that location

A new `-jurl`/`-U`  command line flag was introduced because the already existing `J` flag supports only filesystem paths, where the `:` character (commonly used in URLs) is treated as a separator for multiple paths.

Closes #182

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -46,7 +46,7 @@ import (
 const (
 	flagVerbose    = "verbose"
 	flagJpath      = "jpath"
-	flagSUrl       = "surl"
+	flagJUrl       = "jurl"
 	flagExtVar     = "ext-str"
 	flagExtVarFile = "ext-str-file"
 	flagTlaVar     = "tla-str"
@@ -61,7 +61,7 @@ var overrides clientcmd.ConfigOverrides
 func init() {
 	RootCmd.PersistentFlags().CountP(flagVerbose, "v", "Increase verbosity. May be given multiple times.")
 	RootCmd.PersistentFlags().StringP(flagJpath, "J", "", "Additional jsonnet library search path")
-	RootCmd.PersistentFlags().StringSliceP(flagSUrl, "U", nil, "Additional jsonnet library search path given as an URL")
+	RootCmd.PersistentFlags().StringSliceP(flagJUrl, "U", nil, "Additional jsonnet library search path given as a URL")
 	RootCmd.PersistentFlags().StringSliceP(flagExtVar, "V", nil, "Values of external variables")
 	RootCmd.PersistentFlags().StringSlice(flagExtVarFile, nil, "Read external variable from a file")
 	RootCmd.PersistentFlags().StringSliceP(flagTlaVar, "A", nil, "Values of top level arguments")
@@ -198,7 +198,7 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 		}
 	}
 
-	sURLs, err := flags.GetStringSlice(flagSUrl)
+	sURLs, err := flags.GetStringSlice(flagJUrl)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,6 +45,7 @@ import (
 const (
 	flagVerbose    = "verbose"
 	flagJpath      = "jpath"
+	flagSUrl       = "surl"
 	flagExtVar     = "ext-str"
 	flagExtVarFile = "ext-str-file"
 	flagTlaVar     = "tla-str"
@@ -59,6 +60,7 @@ var overrides clientcmd.ConfigOverrides
 func init() {
 	RootCmd.PersistentFlags().CountP(flagVerbose, "v", "Increase verbosity. May be given multiple times.")
 	RootCmd.PersistentFlags().StringP(flagJpath, "J", "", "Additional jsonnet library search path")
+	RootCmd.PersistentFlags().StringSliceP(flagSUrl, "U", nil, "Additional jsonnet library search path given as an URL")
 	RootCmd.PersistentFlags().StringSliceP(flagExtVar, "V", nil, "Values of external variables")
 	RootCmd.PersistentFlags().StringSlice(flagExtVarFile, nil, "Read external variable from a file")
 	RootCmd.PersistentFlags().StringSliceP(flagTlaVar, "A", nil, "Values of top level arguments")
@@ -186,7 +188,26 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 		searchPaths = append(searchPaths, p)
 	}
 
-	vm.Importer(&jsonnet.FileImporter{JPaths: searchPaths})
+	sURLs, err := flags.GetStringSlice(flagSUrl)
+	if err != nil {
+		return nil, err
+	}
+	for _, u := range sURLs {
+		log.Debugln("Adding jsonnet search path given as URL", u)
+		searchPaths = append(searchPaths, u)
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	importer, err := utils.MakeUniversalImporter(wd, searchPaths)
+	if err != nil {
+		return nil, err
+	}
+
+	vm.Importer(importer)
 
 	extvars, err := flags.GetStringSlice(flagExtVar)
 	if err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -207,6 +207,9 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 		if err != nil {
 			return nil, err
 		}
+		if u.Path[len(u.Path)-1] != '/' {
+			u.Path = u.Path + "/"
+		}
 		searchUrls = append(searchUrls, u)
 	}
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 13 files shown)


petr-k:
@anguslees ping :)

utils/importer.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+			urls = append(urls, *u)
+		} else {
+			if filepath.IsAbs(p) {
+				urls = append(urls, url.URL{Scheme: "file", Path: p})
+			} else {
+				urls = append(urls, joinURL(workDirURL, p))
+			}
+		}
+	}
```
anguslees: If the initial args to this function are already URLs, then this all goes away (or is moved into the caller).  I think the caller case works out simpler, because we already know from the context whether it's a file path or fully specified URL.

utils/importer.go
```diff
@@ -0,0 +1,174 @@
...
+func (importer *universalImporter) Import(dir, importedPath string) (*jsonnet.ImportedData, error) {
+	candidateURLs, err := importer.expandImportToCandidateURLs(dir, importedPath)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get candidate URLs for when importing %s (import dir is %s)", importedPath, dir)
+	}
+
+	var tried []string
+	for _, u := range candidateURLs {
+		tried = append(tried, u.String())
+		importedData := importer.tryImport(u)
```
anguslees: I feel like we should only keep searching on "not found" errors (other errors should be raised immediately) - what do you think?

utils/importer.go
```diff
@@ -0,0 +1,174 @@
...
+					Content:   string(bodyBytes),
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (importer *universalImporter) expandImportToCandidateURLs(dir, importedPath string) ([]url.URL, error) {
+	importedPathURL, err := url.Parse(importedPath)
```
anguslees: This can be made much simpler, thanks to `URL.ResolveReference`:

```go
importedPathURL, err := url.Parse(importedPath)
if err != nil {
       return nil, fmt.Errorf("Import path %q is not valid: %s", importedPath, err)
}

importDirURL, err := url.Parse(dir)
if err != nil {
       return nil, fmt.Errorf("Invalid import dir %q: %s", dir, err)
}

var candidateURLs []*url.URL
candidateURLs = append(candidateURLs, importDirURL.ResolveReference(importedPathURL))

for _, searchURL := range importer.BaseSearchURLs {
￼	candidateURLs = append(candidateURLs, searchURL.ResolveReference(importedPathURL))
}
￼
return candidateURLs, nil
```

utils/importer.go
```diff
@@ -0,0 +1,174 @@
...
+
+	newURL := url
+	newURL.Path = path.Join(paths...)
+	return newURL
+}
+
+/*
+	Converts a dir value to URL, while fixing up the dir if necessary.
+	The `dir` value that jsonnet calls importer's Import function with is transformed in process - the path
+	is cleaned up, which means that consecutive slashes are compacted into one. That poses a problem where
```
anguslees: Oh :(  This seems like a go-jsonnet bug.  The FoundHere value should be an opaque string that is used for error reporting, and passing back to later imports - it shouldn't be assuming it's an actual file path (and by the sounds of it, calling filepath.Clean on it)

Since we're on good terms with the go-jsonnet authors :) I think we should probably just fix that upstream, rather than work around it with regex guesswork.

petr-k: It actually calls `path.Dir()` on the path (see https://github.com/google/go-jsonnet/blob/master/interpreter.go#L415) which internally cleans up the path. Looks like it could be solved just by calling `path.Split()` instead. I'll submit a PR in the `jsonnet-go` repo.

utils/importer.go
petr-k: I agree that the `joinURL` function can be ditched in favor of `URL.ResolveReference()`, but I don't see how it can fully replace the logic here. I believe what you're proposing does not address the following:
* absolute filesystem paths in `importedPath` (because e.g. `/path/to/abc` URL is not an absolute URL, while `file:///path/to/abc` is)
* for relative `dir`s, we need to search in relative to the workdir

Perhaps I'm missing the point - could you please elaborate?

utils/importer.go
petr-k: Agreed.

utils/importer.go
petr-k: The importer's workdir is a URL, as well as the list of search URLs:
```
type universalImporter struct {
	WorkDirURL     url.URL
	BaseSearchURLs []url.URL
	HTTPClient     *http.Client
}
```

`MakeUniversalImporter` is meant as a convenience factory function. It essentially performs the tasks a typical caller would need to do itself (parsing, validation). I could move this somewhere else (`root.go`?). Or did you have something else in mind altogether?

utils/importer.go
anguslees: Oh right, I was assuming my other suggestion [above](https://github.com/ksonnet/kubecfg/pull/187/files#r168891027) - in that version, I just added the workdir to the importer search path.

Absolute paths is an interesting one.  What do we _want_ to do, when someone uses `import "/full/path"`?  I was assuming we still wanted it to be schema-relative (ie: that import from `http://host/foo/bar.jsonnet` would resolve to `http://host/full/path`).

I can think of some contrived examples where I would want (eg) http-hosted jsonnet files to import other files from an absolute path elsewhere in the http server's URL space.  Do you have something that needs `/full/path` to assume `file://`?

Edit: oh, you make another good point however.  If `importedPathURL` is absolute, then we should skip the rest of this function, else we're just going to end up with a list of repeated duplicate `importedPathURL` values...

utils/importer.go
anguslees: I agree it's just moving code around, so I'm not that fussed.

I have a feeling that at some point we're going to want to specify the workdir as a remote http:// URL for some exotic reason (eg: no local files at all), so I was just anticipating that flexibility now.  I agree that it's a small detail, since we can reuse `universalImporter` with another factory if required.

For the search paths, if we construct them as URLs directly at command-line-arg handling time (ie: in the caller of this function), then we *know* whether they are meant to be URLs or local file paths, so we can avoid some of the `IsAbs()` guesswork.

cmd/root.go
```diff
@@ -186,7 +188,26 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
...
+	}
+	for _, u := range sURLs {
+		log.Debugln("Adding jsonnet search path given as URL", u)
+		searchPaths = append(searchPaths, u)
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
anguslees: .. actually, I think we can get rid of the workdir arg entirely, if we just add it to the start of the searchUrls list (?)

utils/importer.go
petr-k: The `import "/full/path"` is indeed interesting. The rationale behind the decision to treat that as a filesystem path was: if I don't know anything about the URLs behind the scenes, imports starting with `/` look like root-based fs path to me, just as imports starting with e.g. `c:\` do on Windows.

But I am on the fence about this.

cmd/root.go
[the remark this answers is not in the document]
petr-k: Actually, the implementation tries to resolve _everything_ that's not an absolute path against the workdir (https://github.com/ksonnet/kubecfg/pull/187/files#diff-1facf542b1747a277b362a0f7cbe6b66R131). Is that not the behavior we want?

utils/importer.go
petr-k: Right, I'll try to simplify this by passing search paths as URLs.

As far as remote workdir goes, I have a little bit different suggestion in mind. Right now, the initial file passed to `kubecfg` is read with `ioutil.ReadFile(path)` which means it's always a filesystem path, possibly relative to the current workdir. What if we just read the initial file using the importer instead? We could then easily support remote files as well as the current behavior.
Note that keeping the filesystem workdir around might still be useful - the initial remote file could contain imports such as `mylib/b.jsonnetlib` that would be searched for relative to the workdir, if not found relative to the remote.

petr-k:
Angus, could you please sum up what you would like to be changed or mark it in the comments? At this point I'm having trouble following all the edits ;)

utils/importer.go
anguslees: (For the archives, the PR was google/go-jsonnet#191)

anguslees:
> Angus, could you please sum up what you would like to be changed or mark it in the comments?

Yep :P  See https://github.com/petr-k/kubecfg/pull/1 for most of the substantial bits.

utils/importer.go
anguslees: If we don't have a specific reason to choose one interpretation of `import "/full/path"` over the other, I suggest we go with schema-relative - just because the code is simpler (it's just a call to ResolveReference, without a special case).  The all-local-files case will be identical to legacy jsonnet import behaviour in either case, which I think is the only true requirement here.

petr-k:
@anguslees So, I did some further little tweaks, check them out. I feel we're getting close to merging this in, what do you think?

petr-k:
@anguslees nothing major that I know of. Changes to the `Importer` interface that are being discussed are not likely to come right now, so I'd say we can go ahead and merge.

petr-k:
@anguslees could you please do a new release with this feature included?

anguslees:
> could you please do a new release with this feature included?

I was drafting the release note, and realised we didn't support `$http_proxy` and friends. I'll make a release once we merge #207

anguslees:
> could you please do a new release with this feature included?

Done!  https://github.com/ksonnet/kubecfg/releases/tag/v0.8.0

PR #190, opened 2018-02-16
Make GC work also with RBAC

### LINKED ISSUE #189: --gc-tag feature doesn't work if it cannot list all the things
I'm using `kubecfg update foo.jsonnet --gc-tag=bar` in a cluster with RBAC enabled and with an account that can create and list (almost) all objects in a few namespaces, but it cannot list many resource types in most of the other namespaces (e.g. secrets).

In this scenario the kubecfg GC feature cannot be used at all because it fails hard when it tries to list all resources in all namespaces (by using `metav1.NamespaceAll`)

```
apiservices.apiregistration.k8s.io is forbidden: User "system:serviceaccount:toolchain-web:toolchain-web-svcaccount" cannot list apiservices.apiregistration.k8s.io at the cluster scope
```

(The same kind of error happens when using e.g. `kubectl get secret --all-namespaces`)

My deployment spans a handful of namespaces; the jsonnet resource specify in which namespaces each object goes.

I'd like to be able to use the GC in this scenario.

mkmik:
Currently the GC does a pass through all the resource types
and performs a single cluster scoped list of all the objects.

This kind of requests is forbidden if the user has no rights list
every object of that type in all the namespaces.
When RBAC is configured, the user is likely to have at least one
namespace + resource type combination that they cannot list.

I tested this on a cluster with 4k objects in 12 namespaces,
and it was 2x slower on my slow ADSL (2m vs 1m before this patch).
That's bearable

But another cluster I have to use has 67 namespaces and it took
9m. That's tough.
There is no significant changes if by parallelizing by namespace.

Hence this change makes this new behavior optional and driven by flags:

1. --gc-list-mode: governs whether to use the current "cluster-scope",
   or the slower "per-namespace" mode.
2. --gc-ns-selector: allows users to only scan namespaces that match a
   given label selector. This is usually going to be some kind of team
   or project label that all namespaces associated with that user would
   have.

Unfortunately we cannot assume kubecfg can set the --gc-tag on the
namespaces objects it operates into, since they might not be writable by
users who have otherwise full control to all the other reasurces in it.

Closes #189

(also removes some dead code)

This PR doesn't add any tests. The existing code path for listing all resources for GC doesn't seem to have many tests either, so I'll shave this yak asynchronously.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -16,31 +16,21 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/ksonnet/kubecfg/pkg/kubecfg"
 )
 
 const (
-	flagCreate = "create"
-	flagSkipGc = "skip-gc"
-	flagGcTag  = "gc-tag"
-	flagDryRun = "dry-run"
-
-	// AnnotationGcTag annotation that triggers
-	// garbage collection. Objects with value equal to
-	// command-line flag that are *not* in config will be deleted.
-	AnnotationGcTag = "kubecfg.ksonnet.io/garbage-collect-tag"
-
-	// AnnotationGcStrategy controls gc logic.  Current values:
-	// `auto` (default if absent) - do garbage collection
-	// `ignore` - never garbage collect this object
-	AnnotationGcStrategy = "kubecfg.ksonnet.io/garbage-collect-strategy"
-
-	// GcStrategyAuto is the default automatic gc logic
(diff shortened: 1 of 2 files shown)


anguslees:
fyi, I added some tests for garbage collection just recently in #193

anguslees:
I like the change generally.  I don't like adding new flags to force onto the user what could be determined automatically.

I think we should either just always do per-namespace listing, or (better) try global and fall back to per-namespace if we get a permission error.

mmikulicic:
> I don't like adding new flags to force onto the user what could be determined automatically.

yeah it makes sense

Issue #146, opened 2017-09-21
`validate` should support (ie: not reject) TPRs/CRDs


anguslees:
There is an upcoming kubernetes 1.8+feature-flag feature that allows CRDs to publish an accompanying openapi schema.  Just-merged docs: https://github.com/kubernetes/kubernetes.github.io/pull/5290

Before this is widely supported and used with CRDs, and with all TPRs, `kubecfg validate` will be unable to fetch the accompanying schema.  Currently this causes kubecfg to fail the input validation with an error, which is not at all useful in config that uses at least one TPR/CRD.

- [x] Add a non-default `validate --ignore-unknown` flag that reduces this to a non-fatal warning
- [ ] Special-case TPRs and CRDs (with undeclared schemas) by default and validate them against a minimal metadata-only schema.

anguslees:
It's not quite as described in the original comment, but I think we've met the goal of this issue.  CRDs are now correctly accepted by default, with their apiVersion/kind validated against discovery info, but with no schema check (at least until the k8s server starts aggregating schema server-side).

PR #191, opened 2018-02-20
Add --ignore-unknown flag to validate

mmikulicic:
that allows kubecfg validate deal with CRDs until we implement a better
fix.

See #146

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -21,19 +21,30 @@ import (
 	"github.com/ksonnet/kubecfg/pkg/kubecfg"
 )
 
+const (
+	flagIgnoreUnknown = "ignore-unknown"
+)
+
 func init() {
 	RootCmd.AddCommand(validateCmd)
+	validateCmd.PersistentFlags().Bool(flagIgnoreUnknown, false, "Don't fail if the schema for a given resource type cannot be fetched")
 }
 
 var validateCmd = &cobra.Command{
 	Use:   "validate",
 	Short: "Compare generated manifest against server OpenAPI spec",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		flags := cmd.Flags()
 		var err error
 
 		c := kubecfg.ValidateCmd{}
 
 		_, c.Discovery, err = restClientPool(cmd)
+		if err != nil {
+			return err
+		}
+
+		c.IgnoreUnknown, err = flags.GetBool(flagIgnoreUnknown)
 		if err != nil {
 			return err
 		}
--- a/pkg/kubecfg/validate.go
+++ b/pkg/kubecfg/validate.go
@@ -35,28 +35,12 @@ type ValidateCmd struct {
 	IgnoreUnknown bool
 }
 
-// newGroupVersionChecker returns a predicate that returns true if the group version is known to the server.
-func newGroupVersionChecker(d discovery.ServerGroupsInterface) (func(schema.GroupVersion) bool, error) {
-	groupList, err := d.ServerGroups()
-	if err != nil {
-		return nil, err
-	}
-
-	groupVersions := sets.NewString()
-	for _, gv := range v1.ExtractGroupVersions(groupList) {
-		groupVersions.Insert(gv.String())
-	}
-	
-	return func(gv schema.GroupVersion) bool {
-		return groupVersions.Has(gv.String())
-	}, nil
-}
-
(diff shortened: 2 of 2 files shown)


cmd/validate.go
```diff
@@ -21,14 +21,20 @@ import (
 	"github.com/ksonnet/kubecfg/pkg/kubecfg"
 )
 
+const (
+	flagIgnoreUnknown = "ignore-unknown"
+)
+
 func init() {
 	RootCmd.AddCommand(validateCmd)
+	validateCmd.PersistentFlags().Bool(flagIgnoreUnknown, false, "Don't fail if the schema for a given resource type cannot be fetched")
```
anguslees: Default to `true` - I'd like kubecfg to just-work without flags.

pkg/kubecfg/validate.go
```diff
@@ -42,7 +43,12 @@ func (c ValidateCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer)
 
 		schema, err := utils.NewSwaggerSchemaFor(c.Discovery, obj.GroupVersionKind().GroupVersion())
 		if err != nil {
-			allErrs = append(allErrs, fmt.Errorf("Unable to fetch schema: %v", err))
+			e := fmt.Errorf("Unable to fetch schema: %v", err)
+			if c.IgnoreUnknown {
+				log.Warnf("%v", e)
```
anguslees: I _think_ we want to drop this to Info, but I'm open to other opinions.

My thinking is that this message will be "normal" for anyone with CRDs (or other "legitimate" missing schemas), and we shouldn't be showing warnings as part of normal.

pkg/kubecfg/validate.go
```diff
@@ -42,7 +43,12 @@ func (c ValidateCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer)
 
 		schema, err := utils.NewSwaggerSchemaFor(c.Discovery, obj.GroupVersionKind().GroupVersion())
 		if err != nil {
-			allErrs = append(allErrs, fmt.Errorf("Unable to fetch schema: %v", err))
+			e := fmt.Errorf("Unable to fetch schema: %v", err)
+			if c.IgnoreUnknown {
```
anguslees: Can we make this more specific to just "not found"?  Eg: I don't want to silently skip _all_ validation if the client can't authenticate to the apiserver or some other unrelated (global) condition.

mmikulicic: done, but instead of checking the error, I check whether the GroupVersion is known (incidentally, this is the same check that is performed later on by `util.NewSwaggerSchemaFor`)

pkg/kubecfg/validate.go
mmikulicic: see above (until we actually check for CRDs, possibly in another PR, this should stay a warning)

cmd/validate.go
mmikulicic: I wouldn't mind defaulting this to true once it's correct (i.e. safe), especially w.r.t only skipping errors caused by CRDs (without schema) rather than any generic error which can possibly be caused by a typo in the resource type. Not everybody is using CRDs, and those who are not might expect kubecfg validate to catch typos in resource types.

I'm willing to follow through this task (#146) and properly implement everything, but I wanted to do it incrementally since right now we can't use kubecfg validate at all due to CRDs.

mmikulicic:
@anguslees, friendly ping (in "outdated")
(github code review feature is confusing, it hides my answers)

pkg/kubecfg/validate.go
```diff
@@ -20,27 +20,59 @@ import (
...
+
+// newGroupVersionChecker returns a predicate that returns true if the group version is known to the server.
+func newGroupVersionChecker(d discovery.ServerGroupsInterface) (func(schema.GroupVersion) bool, error) {
+	groupList, err := d.ServerGroups()
+	if err != nil {
+		return nil, err
+	}
+	groupVersions := v1.ExtractGroupVersions(groupList)
+
+	return func(gv schema.GroupVersion) bool {
```
anguslees: If we're going to do repeated presence tests, we may as well just build/keep a `k8s.io/apimachinery/pkg/util/sets.String` of the `gv.String()`s - indeed you can probably just return that and have the caller test for presence in the set directly (without the helper closure).

Issue #194, opened 2018-02-27
kubecfg uses obsolete swaggerapi, removed in client-go v6.0.0


prydonius:
We cannot easily update kubecfg to use the latest client-go release (v6.0.0) as they have removed the Swagger 1.2 retriever functions, which we rely on for validation.

See https://github.com/kubernetes/client-go/blob/master/CHANGELOG.md#v600 and https://github.com/kubernetes/kubernetes/pull/53441 for more info.

In order to update, we will need to reimplement where we're using Swagger with the OpenAPI functions.

cc @anguslees @ngtuna

anguslees:
This was fixed in #215

PR #195, opened 2018-02-27
update to client-go v5.0.1

prydonius:
See #194 as to why we cannot yet upgrade to v6.0.0


integration/integration_suite_test.go
```diff
@@ -25,9 +25,6 @@ import (
 
 	// For client auth plugins
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
-	// For apimachinery serialisation magic
-	_ "k8s.io/client-go/pkg/api/install"
```
prydonius: I'm not sure where this went/why it's needed. @anguslees?

PR #197, opened 2018-03-07
Embed kubecfg.libsonnet inside the kubecfg binary

anguslees:
This change adds a new "internal:///" URL to the end of the search
path, based on files embedded in the binary using go-bindata.
Currently this only includes `kubecfg.libsonnet`.

Note the new generated `utils/bindata.go` is added to git so a naive
"go get ..." is still possible.  Run `make generate` to regenerate
this file when needed (and include any changes in PR).


anguslees:
(cc @petr-k - look at what other cleanups your URL change allows!)

anguslees:
> to force it to use embedded lib

Oh, interesting point.. Since it's a test file, I kind of _want_ it to use the kubecfg.jsonnet in the same directory, and I hadn't thought about the ambiguity vs the internal version.  The way the Makefile/unittest works, it will always use the locally built kubecfg when running the test (in the clean travis CI environment), so I think we're safe either way.

I'm going to merge this as-is, but it's definitely something to keep in mind if this ever confuses anyone.
  (Perhaps I should change it to `import "./kubecfg.libsonnet"` ?)

PR #198, opened 2018-03-07
Support multiple `-J` args

anguslees:
Previously only a single colon-separated `-J` arg was accepted, and
treated like `KUBECFG_JPATH`.  This was a bit weird and hard to use.

This change allows `-J` to be given multiple times, each time being a
simple complete path (no internal separator) - as it should always
have been.  For similar reasons, `-U` is also changed from a
`StringSlice` to a `StringArray`.


anguslees:
(fyi @petr-k)

PR #199, opened 2018-03-08
Improve handling of missing schema

mmikulicic:
This change gets rid of the `--ignore-unknown` flag.
It skips checking the schema if the resource is known but
the schema is missing.

See #146

### DIFF
--- a/pkg/kubecfg/validate.go
+++ b/pkg/kubecfg/validate.go
@@ -20,7 +20,7 @@ import (
 	"io"
 
 	log "github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


pkg/kubecfg/validate.go
```diff
@@ -30,33 +30,40 @@ import (
...
-		if c.IgnoreUnknown && !groupVersions.Has(gv.String()) {
-			log.Warnf("Skipping validation of %s because schema for %s is unknown", desc, gv)
-			continue
-		}
 
 		var allErrs []error
 
 		schema, err := utils.NewSwaggerSchemaFor(c.Discovery, gv)
 		if err != nil {
+			if _, known := knownResources[gv.String()][resName]; errors.IsNotFound(err) && known {
```
anguslees: If the top-level entry doesn't exist (`gv.String()` is unrecognised), this will explode - correct?
If so - the above suggestion to flatten this map will make that go away...

pkg/kubecfg/validate.go
```diff
@@ -30,33 +30,40 @@ import (
...
+				log.Warnf("Skipping validation of known resource %s %s that lacks a registered schema", gv, resName)
```
anguslees: I think we should drop this to an Info now that we know it's a valid resource type, and we've done the best we can.

pkg/kubecfg/validate.go
```diff
@@ -30,33 +30,40 @@ import (
...
+
+	serverResourceList, err := c.Discovery.ServerResources()
 	if err != nil {
 		return err
 	}
-	groupVersions := sets.NewString(v1.ExtractGroupVersions(groupList)...)
+	for _, rl := range serverResourceList {
+		knownResources[rl.GroupVersion] = sets.String{}
+		for _, r := range rl.APIResources {
+			knownResources[rl.GroupVersion][r.Name] = sets.Empty{}
```
anguslees: Hrm.  Since we're going from actual objects here, I think the "known" set should be based on GroupVersionKind, not GroupVersionResource - that we we're avoiding a potentially confusing attempt at mapping kind->resource (`utils.ResourceNameFor`) for unknown kinds.

anguslees:
> but would be nice to add a few simple tests for this sometime.

Agreed!  It really needs to be an integration test, because much of the subtlety here (like this PR) depends on specific k8s server-side behaviour.

pkg/kubecfg/validate.go
mmikulicic: in Go it's perfectly legal to read from a nil map. 

https://golang.org/ref/spec#Map_types:
> A nil map is equivalent to an empty map except that no elements may be added.

so it does work as expected (I did even try with resources with bad group and bad resource names):
https://play.golang.org/p/8IhkgjN8VPM

That said, I agree that a flat map/set is better.

PR #201, opened 2018-03-14
Improve "failed to read ~/.kube/config" error message

anguslees:
Trivial change to add more context to the "error while reading
$KUBECONFIG" error message.

I was recently confused by the non-specificity of the error message
that occurs in this situation, and if _I'm_ confused, then I feel bad
for other users :/

Before:
```console
 % KUBECONFIG=/noexist ./kubecfg update
ERROR invalid configuration: no configuration has been provided
```

After:
```console
 % KUBECONFIG=/noexist kubecfg update
ERROR Unable to read kubectl config: invalid configuration: no configuration has been provided
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -387,7 +387,7 @@ func dumpJSON(v interface{}) string {
 func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.DiscoveryInterface, error) {
 	conf, err := clientConfig.ClientConfig()
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("Unable to read kubectl config: %v", err)
 	}
 
 	disco, err := discovery.NewDiscoveryClientForConfig(conf)


cmd/root.go
```diff
@@ -387,7 +387,7 @@ func dumpJSON(v interface{}) string {
 func restClientPool(cmd *cobra.Command) (dynamic.ClientPool, discovery.DiscoveryInterface, error) {
 	conf, err := clientConfig.ClientConfig()
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("Unable to read kubectl config: %v", err)
```
mmikulicic: usually errors shouldn't be capitalized (https://github.com/golang/go/wiki/CodeReviewComments#error-strings), but I see that we're printing this out directly so we might be tempted to do it

anguslees: oh interesting.  I knew k8s pushed this, but didn't realise it was general golang style.  I will have to adapt my habits.

(And indeed, I've always capitalised after a colon in english, except for trivial lists.  Now I learn that's debatable too...)

PR #206, opened 2018-03-28
[jjo] fix int64 diff handling

jjo:


jjo:
Without the `int64` case, test would fail as per below details
(fyi found this while working on some actual `Deployment` diffs):
```
$WORK/github.com/ksonnet/kubecfg/pkg/kubecfg/_test/kubecfg.test -test.v=true -test.run=TestRemoveFields
=== RUN   TestRemoveFields
--- FAIL: TestRemoveFields (0.00s)
panic: Found unexpected type int in json unmarshal (value=10) [recovered]
        panic: Found unexpected type int in json unmarshal (value=10)
```

PR #207, opened 2018-03-29
Honour `$http_proxy` et al in HTTP Transport

anguslees:
Use (a copy of) `http.DefaultTransport` instead of just a bare
`http.Transport` for HTTP/HTTPS requests.  In particular, the new
Transport honours the usual `http_proxy`, etc environment variables,
and has finite connection timeouts.

### DIFF
--- a/utils/importer.go
+++ b/utils/importer.go
@@ -4,10 +4,12 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/elazarl/go-bindata-assetfs"
 	jsonnet "github.com/google/go-jsonnet"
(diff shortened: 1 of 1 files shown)


utils/importer.go
```diff
@@ -58,7 +60,21 @@ A real-world example:
 	and downloaded from that location
 */
 func MakeUniversalImporter(searchUrls []*url.URL) jsonnet.Importer {
-	t := &http.Transport{}
+	// Reconstructed copy of http.DefaultTransport (to avoid
+	// modifying the default)
+	t := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
```
mmikulicic: the DialContext function can be taken defaultTransport

you _could_ reuse the DialContext function, but that assumes DefaultTransport will forever be a Transport, yet the global var's type is http.RoundTripper.

My suggestion would be to just define the transport you want to use fully and not to rely on the DefaultTransport value, which is provided for convenience only.

PR #210, opened 2018-04-16
Parse server GitVersion

### LINKED ISSUE #200: kubecfg should be more robust against bad server Major/Minor info
hi all

step 1， install the minikube and start
step 2.   download the release binary
step 3.   using the configuration under example and run kubecfg create, it'ok
step 4   trying to delete by running kubecfg delete, it failed like below

get error like below
`ERROR strconv.Atoi: parsing "": invalid syntax`

```
~/dev/kubecfg$ kubectl version
Client Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.4", GitCommit:"793658f2d7ca7f064d2bdf606519f9fe1229c381", GitTreeState:"clean", BuildDate:"2017-08-17T17:03:51Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"", Minor:"", GitVersion:"v1.9.0", GitCommit:"925c127ec6b946659ad0fd596fa959be43f0cc05", GitTreeState:"clean", BuildDate:"2018-01-26T19:04:38Z", GoVersion:"go1.9.1", Compiler:"gc", Platform:"linux/amd64"}
```

```
 :~/dev/kubecfg$ ./kubecfg version
kubecfg version: v0.7.2
jsonnet version: v0.9.5
client-go version: v1.7.3-beta.0+$Format:%h$
(py36) yongweixing@ :~/dev/kubecfg$ ./kubecfg version
kubecfg version: v0.7.2
jsonnet version: v0.9.5
client-go version: v1.7.3-beta.0+$Format:%h$
```

```
 :~/dev/kubecfg$ minikube version
minikube version: v0.25.0
```

andresmgot:
Fixes #200 

In case that server version `Major` and `Minor` are not set or cannot be parsed use `GitVersion`. If all of the previous approaches fail use a default version. Using 1.8 as default version to match `client-go` current version.

### DIFF
--- a/utils/meta.go
+++ b/utils/meta.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -18,21 +19,39 @@ type ServerVersion struct {
 	Minor int
 }
 
-// ParseVersion parses version.Info into a ServerVersion struct
-func ParseVersion(v *version.Info) (ret ServerVersion, err error) {
-	ret.Major, err = strconv.Atoi(v.Major)
-	if err != nil {
-		return
+func parseGitVersion(gitVersion string) (ServerVersion, error) {
+	// Format v0.0.0(-master+$Format:%h$)
+	re := regexp.MustCompile("v([0-9])+.([0-9])+.[0-9]+-?.*")
+	parsedVersion := re.FindStringSubmatch(gitVersion)
+	if len(parsedVersion) != 3 {
+		return ServerVersion{}, fmt.Errorf("Unable to parse git version %s", gitVersion)
 	}
+	var ret ServerVersion
+	var err error
+	ret.Major, err = strconv.Atoi(parsedVersion[1])
+	if err != nil {
+		return ServerVersion{}, err
+	}
+	ret.Minor, err = strconv.Atoi(parsedVersion[2])
+	if err != nil {
+		return ServerVersion{}, err
+	}
+	return ret, nil
+}
 
+// ParseVersion parses version.Info into a ServerVersion struct
+func ParseVersion(v *version.Info) (ServerVersion, error) {
+	var ret ServerVersion
+	var err error
+	ret.Major, err = strconv.Atoi(v.Major)
 	// trim "+" in minor version (happened on GKE)
 	v.Minor = strings.TrimSuffix(v.Minor, "+")
-
 	ret.Minor, err = strconv.Atoi(v.Minor)
 	if err != nil {
-		return
+		// Try to parse using GitVersion
+		return parseGitVersion(v.GitVersion)
 	}
-	return
+	return ret, err
 }
 
 // FetchVersion fetches version information from discovery client, and parses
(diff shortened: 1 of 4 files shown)


utils/meta.go
```diff
@@ -18,21 +19,39 @@ type ServerVersion struct {
...
+		return ServerVersion{}, err
 	}
+	return ret, nil
+}
 
+// ParseVersion parses version.Info into a ServerVersion struct
+func ParseVersion(v *version.Info) (ServerVersion, error) {
+	var ret ServerVersion
+	var err error
+	ret.Major, err = strconv.Atoi(v.Major)
```
arapulido: Shouldn't we check this error as well and return parseGitVersion if it is not nil? If not, the minor one below will overwrite this one.

andresmgot: true, missed the case in which `Major` is not correct and `Minor` is

utils/meta.go
```diff
@@ -18,21 +19,44 @@ type ServerVersion struct {
 	Minor int
 }
 
+func parseGitVersion(gitVersion string) (ServerVersion, error) {
+	// Format v0.0.0(-master+$Format:%h$)
+	re := regexp.MustCompile("v([0-9])+.([0-9])+.[0-9]+-?.*")
```
anguslees: `MustCompile` panics if it doesn't compile - so is best used somewhere where it is _always_ invoked (and thus panics early enough that the developer _has_ to notice their error).

I suggest moving this up to a global:
```go
var gitVersionRe := regexp.MustCompile(...)

func parseGitVersion(...) {
  ...
}
```

anguslees: What's the trailing `-?` for?  Since `+` is a greedy operator by default, I think this regex is equivalent to just `...[0-9]+.*`

utils/meta.go
```diff
@@ -18,21 +19,44 @@ type ServerVersion struct {
...
+	if len(parsedVersion) != 3 {
+		return ServerVersion{}, fmt.Errorf("Unable to parse git version %s", gitVersion)
+	}
+	var ret ServerVersion
+	var err error
+	ret.Major, err = strconv.Atoi(parsedVersion[1])
+	if err != nil {
+		return ServerVersion{}, err
+	}
+	ret.Minor, err = strconv.Atoi(parsedVersion[2])
```
anguslees: Can we still get a trailing + here, or does that only appear in the "real" `version.Minor` and not the gitVersion?  (I have no idea.  Perhaps look at a GKE version?)

utils/meta.go
```diff
@@ -44,6 +68,11 @@ func FetchVersion(v discovery.ServerVersionInterface) (ret ServerVersion, err er
 	return ParseVersion(version)
 }
 
+// GetDefaultVersion returns a default server version (1.8)
```
anguslees: Add a comment here to make it clear that this fallback value is some time-sensitive guess and should not be assumed to stay the same forever.  Perhaps:
"This value will be updated periodically to match a current/popular version corresponding to the age of this code"

utils/meta.go
andresmgot: it is, yes

utils/meta.go
andresmgot: In theory it is not. The version should be in the format `v1.2.3` (the `+` suffix was a GKE is a way to specify that the git version was `v1.8.8` instead of `v1.8.0`)

utils/meta.go
andresmgot: added

Issue #79, opened 2017-08-14
`kubecfg show` needs better error messages / usefulness at debug-time


anguslees:
`kubecfg show` (and `jsonnet eval`) is basically the only tool for exploring evaluation and debugging errors.  Currently `kubecfg show` is awful for this.

- [ ] Allow `kubecfg show` to work on jsonnet sub-expressions not just whole API objects
- [ ] When walking the jsonnet result, give better (some!) context in `Unexpected object structure` error message
- [ ] Accept arbitrary jsonnet expressions, not just files on the `kubecfg show` command line (like `jsonnet eval`)
- [ ] Consider other debugging UX improvements, like a repl?

PR #214, opened 2018-05-08
Ignore jsonnet `null` results; Add context to jsonWalk error message

anguslees:
When walking the jsonnet result and encountering something that
is *not* a collection of kubernetes objects, the previous error
message was totally useless in helping find the actual error.

Was: `Unexpected object structure`
Now: `Looking for kubernetes object at <top>.foo.bar[1], but instead
found float64`

Also: Silently ignore `null` values encountered during the walk. They
don't exist in the Kubernetes API, are useful to "white out" values in
jsonnet, and have no other ambiguous interpretation.

Also: Avoid serialising to json and back *again* when constructing the
resulting `unstructured.Unstructured` objects.

Partial-fix for #79

### DIFF
--- a/utils/acquire.go
+++ b/utils/acquire.go
@@ -96,15 +96,34 @@ func yamlReader(r io.ReadCloser) ([]runtime.Object, error) {
 	return ret, nil
 }
 
-func jsonWalk(obj interface{}) ([]interface{}, error) {
+type walkContext struct {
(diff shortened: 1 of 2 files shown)


anguslees:
@jjo: fyi, the `null` change here should make parsing yaml streams with empty docs "just work" without having to explicitly filter them out...

PR #215, opened 2018-05-25
Switch to OpenAPI and drop Swagger

mmikulicic:
Addresses #194

Most of the validation now happens within the kube-openapi library, greatly simplifying our code.

I actually developed this in a branch where I bumped up the k8s go-client version to >=6.0.0.
This commit is not upgrading the client though.

that are no longer relevant now, since we use full GroupKind information to fetch
the relevant schema.


mmikulicic:
tests fails on minikube test cluster based on 1.7.0, likely because the openapi schema contains 
entries like `io.k8s.kubernetes.pkg.api.v1.ConfigMap`.

More recent schemas state that:

```
      _ref: "#/definitions/io.k8s.api.core.v1.ConfigMap"
      description: "Deprecated. Please use io.k8s.api.core.v1.ConfigMap instead."
```

mmikulicic:
CC @gus, @jjo

utils/sort.go
```diff
@@ -16,119 +16,63 @@
...
-		return result
+func containsPodSpec(disco discovery.OpenAPISchemaInterface, gvk schema.GroupVersionKind) bool {
+	result, ok := podSpecCache[gvk.String()]
+	if ok {
+		return bool(result)
 	}
 
-	swagger, err := disco.SwaggerSchema(gvk.GroupVersion())
+	oapi, err := NewOpenAPISchemaFor(disco, gvk)
 	if err != nil {
```
anguslees: should still debug log the error here, otherwise we're throwing away that information.

utils/sort.go
```diff
@@ -16,119 +16,63 @@
...
-		return nil
-	}
-	seen.Insert(typeName)
-
-	if err := visitor(typeName); err != nil {
-		return err
-	}
+// a podSpecVisitor traverses a schema tree and records whether the schema
+// contains a PodSpec resource.
+type podSpecVisitor bool
```
anguslees: (Nice!)

utils/client.go
```diff
@@ -142,6 +124,10 @@ func (c *memcachedDiscoveryClient) OpenAPISchema() (*openapi_v2.Document, error)
 	return schema, nil
 }
 
+func (c *memcachedDiscoveryClient) SwaggerSchema(version schema.GroupVersion) (*swagger.ApiDeclaration, error) {
+	return nil, fmt.Errorf("swagger API is deprecated")
```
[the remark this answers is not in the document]
mmikulicic: yes, I wanted to create a single PR that bumps the client version *and* adapts our use of it, but it was impractical.

This method had to be kept in order to make this struct implement discovery.DiscoveryInterface (whose definition in the next version of k8s client-go library gets rid of this method).

Added comment with explicit TODO about removal (and panic instead of Error)

utils/sort.go
mmikulicic: fair enough

PR #220, opened 2018-06-04
Upgrade client-go to kubernetes-1.10.3

mmikulicic:


pkg/kubecfg/update_test.go
```diff
@@ -78,7 +78,7 @@ func TestEligibleForGc(t *testing.T) {
 		if ref.Controller != nil {
 			c["controller"] = *ref.Controller
 		}
-		u.Object["metadata"].(map[string]interface{})["ownerReferences"] = []map[string]interface{}{c}
+		u.Object["metadata"].(map[string]interface{})["ownerReferences"] = []interface{}{c}
```
anguslees: oops.  I wonder why the earlier incorrect structure didn't seem to matter ..

mmikulicic: it did, they changed the internal impl of unstructured.

Ideally we should use the public API to set values here.

anguslees: +1  Although I note the "Unstructured content" representation is also public via `UnstructuredContent()` (and the `Object` property).

PR #223, opened 2018-07-17
Default to --ignore-unknown=true

### LINKED ISSUE #211: kubecfg with custom resources fails validation
Setup: use kubecfg to define a CustomResourceDefinition, and then an object of that type.

`kubecfg update` will fail validation, because right now, the cluster doesn't know about the custom resource, so the 2nd object isn't valid. But if the CRD is applied before the object (which it is, in the current resource ordering), the update will result in fully valid state.

If the server says it doesn't know about a resource type, but the update has a CustomResourceDefinition for that type, kubecfg should relax validation slightly and allow the object to pass validation with limited checking.

anguslees:
Invert the default for `--ignore-unknown` so that we now ignore
unknown types by default, and require `--ignore-unknown=false` to make
these fatal.  This follows the `kubectl` behaviour, and makes the
`kubecfg` default more likely to "just work" in the presence of CRDs,
etc.

Fixes #211

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -27,7 +27,7 @@ const (
 
 func init() {
 	RootCmd.AddCommand(validateCmd)
-	validateCmd.PersistentFlags().Bool(flagIgnoreUnknown, false, "Don't fail if the schema for a given resource type is not found")
+	validateCmd.PersistentFlags().Bool(flagIgnoreUnknown, true, "Don't fail if the schema for a given resource type is not found")
 }
 
 var validateCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


arapulido:
@anguslees great! thanks! very much needed :) Can we create a new release once this is in?

anguslees:
> Can we create a new release once this is in?

Yes, that's the plan for tomorrow :)

anguslees:
> Can we create a new release once this is in?

Done: https://github.com/ksonnet/kubecfg/releases/tag/v0.9.0

PR #225, opened 2018-08-31
Add support for Google Container Registry

burdiyan:
Original resolver implementation in `kubecfg` was not working with protected Google Container Registry repositories. It was failing with 401 Unauthorized response code, even though `docker pull` for the same image is working fine.

This PR changes the RegistryResolver implementation by using a pretty solid external package of container registry utilities: https://github.comgenuinetools/reg.

I removed the `utils/resolver_test.go` file as it is not relevant that much anymore since external package is used.

This issue was a real blocker for us for using `kubecfg`, as we rely very heavily on deployments by the exact image digest.


utils/resolver.go
```diff
@@ -18,8 +18,10 @@ package utils
 import (
 	"bytes"
 	"fmt"
-	"net/http"
-	"strings"
+
+	"github.com/genuinetools/reg/repoutils"
+
```
anguslees: Remove this blank line between the 3rd party imports

anguslees:
Urgh, every time I use the github editor to make "obvious" changes without being able to test them first, I make things worse :/

PR #227, opened 2018-09-07
Update reg to fix image resolution bug

burdiyan:
There was a bug in github.com/genuinetools/reg https://github.com/genuinetools/reg/issues/131
that was causing wrong image digest resolutions. The bug was fixed upstream, so this commit
updates the dependencies to pull this bug fix.


burdiyan:
After I ran this command I cannot build the binary anymore inside Docker container. It's complaining about missing libraries:

```
vendor/golang.org/x/net/idna/idna.go:23:2: cannot find package "golang.org/x/text/secure/bidirule" in any of:
        /go/src/github.com/ksonnet/kubecfg/vendor/golang.org/x/text/secure/bidirule (vendor tree)
        /usr/local/go/src/golang.org/x/text/secure/bidirule (from $GOROOT)
        /go/src/golang.org/x/text/secure/bidirule (from $GOPATH)
vendor/golang.org/x/net/idna/idna.go:24:2: cannot find package "golang.org/x/text/unicode/norm" in any of:
        /go/src/github.com/ksonnet/kubecfg/vendor/golang.org/x/text/unicode/norm (vendor tree)
        /usr/local/go/src/golang.org/x/text/unicode/norm (from $GOROOT)
        /go/src/golang.org/x/text/unicode/norm (from $GOPATH)
vendor/github.com/PuerkitoBio/purell/purell.go:19:2: cannot find package "golang.org/x/text/width" in any of:
        /go/src/github.com/ksonnet/kubecfg/vendor/golang.org/x/text/width (vendor tree)
        /usr/local/go/src/golang.org/x/text/width (from $GOROOT)
        /go/src/golang.org/x/text/width (from $GOPATH)
```

Seems like it removes some of the libraries that are actually used.

anguslees:
> Seems like it removes some of the libraries that are actually used.

Huh, in that case we'd better not remove them :)

Commit as-is, and I'll find out what's up with govendor later.  We probably need to add those libraries to some explicit list to avoid a future `remove +u` breaking things again.

Issue #229, opened 2018-11-06
kubecfg diff is showing incorrect information (marking unchanged lines as changed, ...)


emempi:
When running kubecfg diff on our infrastructure, the output often contains lines that do no change, e.g.:
```
$ kubecfg diff my/config.jsonnet
...
-    "replicas": 1,
+    "replicas": 1,
...
               {
-                "containerPort": 8089,
+                "containerPort": 8089,
                 "name": "whatever",
                 "protocol": "TCP"
               }
...
```

Moreover, we've seen cases when the diff was out-of-context, e.g. environment variable showing a value change to a value that should belong to another environment variable according to the configuration.

```
$ kubecfg version
kubecfg version: v0.9.0
jsonnet version: v0.10.0
client-go version: v0.0.0-master+$Format:%h$
```

seh:
Not having inspected the code, I think these port differences show up due to the “int ot string” representation used in the Kubernetes API. Perhaps _kubecfg_ is reading an integer (or just a number) from the evaluated Jsonnet output, but the API server is presenting a string for the corresponding field’s value.

emempi:
That might indeed be a part of the problem (but IIRC, I tried representing my values as both strings and ints and in both cases I got invalid output, so it might be a bit more intricate). Another issue might be hidden inside gojsondiff library -- there are a few open issues that result in incorrect diffs produced, e.g.:

* https://github.com/yudai/gojsondiff/issues/24
* https://github.com/yudai/gojsondiff/issues/30

W.r.t. the issues I mentioned, while the first one (no-op diff lines) are just an annoyance, the other issue I mentioned (likely related to https://github.com/yudai/gojsondiff/issues/24) is actually dangerous, since it presents the user with false infrastructure changes data; I'd recommend prioritizing that one.

anguslees:
Thanks for the problem report.  Yes, the current diff code is the simplest thing I could put together, and could benefit from some actual thought.  Designs/patches up to entire replacements welcome.

emempi:
Would you be OK with a plain stupid "format both JSONs and run diff on them"-like approach? That would probably be reasonably fool-proof (although without any json-specific heuristics).

emempi:
Friendly ping :) I attached my PR to update the diff code a bit; please let me know what you think about it. Some custom formatting code was needed to maintain backward compatibility.

[...]

PR #230, opened 2018-11-13
Rework kubecfg diff code.

emempi:
The old code relied on gojsondiff library, which provides more semantic
JSON diffs. This library, however, had some long-standing issues which
have not been yet fixed, namely:

* https://github.com/yudai/gojsondiff/issues/24
* https://github.com/yudai/gojsondiff/issues/30

These sometimes resulted in diffs that were too noisy and/or showed
false information; this has been reported as

https://github.com/ksonnet/kubecfg/issues/229

This patch reworks kubecfg's diff code to first marshal JSON using
golang's encoding/json and then calculates the diff using
diffmatchpatch library, formatting the output as close to the
original as possible. Although this results in a more
syntactic-level diff, it might just do the job until gojsondiff
gets the above-mentioned bugs fixed.

Signed-off-by: Milan Plzik <milan.plzik@ceai.io>

### DIFF
--- a/pkg/kubecfg/diff.go
+++ b/pkg/kubecfg/diff.go
@@ -38,6 +38,9 @@ import (
 
 var ErrDiffFound = fmt.Errorf("Differences found.")
 
+// Matches all the line starts on a diff text, which is where we put diff markers and indent
+var DiffLineStart = regexp.MustCompile("(^|\n)(.)")
+
 // DiffCmd represents the diff subcommand
 type DiffCmd struct {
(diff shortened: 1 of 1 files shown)


emempi:
PTAL.

mkmik:
LGTM.

but sadly EMENOMAINTAINER of this repo; pinging my colleague @anguslees.

jjo:
LGTM also, thanks @emempi -- verified it fixes false diff positives
on numeric values, and much better handles e.g. configmap diffs
with multiline valued fields.

PR #235, opened 2019-01-15
Update to go-jsonnet 0.12.1

mkmik:
Relevant changelog (0.10.x -> 0.12.x):

- Fix std.parseHex that rejected certain correct inputs
- Added std.find and std.findSubstr
- Added std.parseJson
- std.native("foo") now returns null if foo doesn't exist, instead of an error. This is useful for falling back on Jsonnet implementations if a faster native one isn't available.
- sort and set functions can now be given a "keyF" function for controlling the ordering
- std.trace for debugging
- std.parseHex
- std.parseOctal
- Major (2x-10x) improvements in performance via a substantial refactoring of the interpreter


utils/importer.go
```diff
@@ -89,52 +89,49 @@ type universalImporter struct {
 	HTTPClient     *http.Client
 }
 
-func (importer *universalImporter) Import(dir, importedPath string) (*jsonnet.ImportedData, error) {
+func (importer *universalImporter) Import(dir, importedPath string) (jsonnet.Contents, string, error) {
```
anguslees: If I remember the upstream jsonnet change correctly, I think we are now expected to do our own caching in the Importer (previously this was done internally in jsonnet).

Can you test importing the same path twice, and see (eg via the following debug print) how many times this gets invoked?   If multiple times, then we need to also add our own simple cache as part of this PR - especially because we support expensive/remote things like URLs.

(For bonus points make a test that verifies we only get called once)

To be completely accurate, we should also read the initial jsonnet file through the cache too.  Atm the initial file is read separately (outside the Importer) - see the TODO in `utils.jsonnetReader`.  I'm ok punting on this aspect for now, since (because of the same TODO) the initial file can only be a local path, which is cheap and unlikely to change between reads.

utils/importer.go
```diff
@@ -81,60 +81,64 @@ func MakeUniversalImporter(searchUrls []*url.URL) jsonnet.Importer {
...
 	if err != nil {
-		return nil, fmt.Errorf("Could not get candidate URLs for when importing %s (import dir is %s)", importedPath, dir)
+		return jsonnet.Contents{}, "", fmt.Errorf("Could not get candidate URLs for when importing %s (import dir is %s)", importedPath, dir)
 	}
 
 	var tried []string
 	for _, u := range candidateURLs {
 		tried = append(tried, u.String())
-		importedData, err := importer.tryImport(u)
+		importedData, foundAt, err := importer.tryImport(u)
```
anguslees: Am I reading this correctly?  I think you are fetching the content every time, and _then_ discarding the content if it is a cache hit.  Also, we may as well cache not-found results, since these might be common with a search path and there is no point trying them again.

I think we want something like:
```golang
for _, u := range candidateURLs {
        c, hit := importer.cache[u.String()]
        if !hit {
                // no entry in cache -> fetch from origin
                var err error
                c, err = importer.tryImport(u)
                if err != errNotFound {
                      return jsonnet.Contents{}, "", err
                }
                importer.cache[u.String()] = c  // maybe nil for not-found
        }
        if c != nil {
                return c, u.String(), nil
        }
        tried = append(tried, u.String())
}
```

Also: After looking at it more closely, I don't think `tryImport` needs to be modified to return an additional `foundAt` result - this function is always called with absolute URLs, and so `foundAt` is always `url.String()`.  It also appears we always want the URL in string form at this point (including inside `tryImport`), so it also might make sense to just do that conversion once somewhere.

testdata/import_test.jsonnet
```diff
@@ -0,0 +1 @@
+(import "imported.jsonnet") + (import "imported.jsonnet")
```
anguslees: Minor: missing trailing newline

testdata/imported.jsonnet
```diff
@@ -0,0 +1 @@
+{}
```
anguslees: Minor: missing trailing newline

utils/importer.go
anguslees: Nice test.

PR #239, opened 2019-02-11
Install tool dependencies separately

mkmik:
Otherwise running `go get` in in go1.11 will update the go.mod file
and causing our build to fail because we check for dirty working dir after generate.

This change also ensures that we pin the version of the tools we use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ addons:
       - ca-certificates
 
 install:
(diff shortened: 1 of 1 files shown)


anguslees:
Tests are failing with:
```console
go-bindata -nometadata -ignore .*_test\.|~$ -pkg utils -o bindata.go -prefix ../ ../lib/...
flag provided but not defined: -nometadata
```

mkmik:
They pass now

PR #243, opened 2019-03-03
Upgrade "client-go" dependency to version "kubernetes-1.13.4"

seh:
In order to accommodate more recent _client-go_ credential plugin API versions, run the following commands to freshen our dependencies.
```shell
go get k8s.io/client-go@kubernetes-1.13.4
go mod tidy
```
Here we target [the _kubernetes-1.13.4_ release](https://github.com/kubernetes/client-go/releases/tag/kubernetes-1.13.4) of _client-go_.

Fixes #240.

PR #244, opened 2019-03-04
Update client-go to 10.0 (aka kubernetes 1.13)

### LINKED ISSUE #240: kubecfg does not work with client-go credential plugins at version "client.authentication.k8s.io/v1beta1"
Using _kubecfg_ version 0.9.1 on macOS, if my KUBECONFIG file uses [a client-go credential plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins), _kubecfg update_ fails with the following message:

> ERROR exec plugin: invalid apiVersion "client.authentication.k8s.io/v1beta1"

That line appears to originate [in the `exec.newAuthenticator` function](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/plugin/pkg/client/auth/exec/exec.go#L121). Looking at the history of that file, I wonder if _kubecfg_'s version of _client-go_ is old enough to predate the introduction of "client.authentication.k8s.io/v1beta1," which came in with this commit on 31 May 2018: [*client-go: promote exec plugin support to beta*](https://github.com/kubernetes/kubernetes/commit/0a9164e73aedd898a535a64fcc992d3e50002d61).

According to [the current _go.mod_ file](https://github.com/ksonnet/kubecfg/blob/master/go.mod#L39), our version of _client-go_ is "v0.0.0-20180521144600-29ae1f00c3d8."

Here is the output of _kubecfg version_:
```
kubecfg version: v0.9.1
jsonnet version: v0.10.0
client-go version: v0.0.0-master+$Format:%h$
```

anguslees:
Update k8s.io/* to kubernetes 1.13.4 (or equivalent git tag).

- Numerous changes to use new upstream dynamic client.
- Also rewrite `utils.ClientForResource` to use standard upstream
  RESTMapper, rather than carry our own equivalent.

Fixes #240


anguslees:
(Thanks to @seh for an earlier iteration in #243)

anguslees:
(This is failing integration test - I've broken the simple yaml-serialisation code it uses.  Will fix tomorrow)

PR #246, opened 2019-03-14
Switch to strategic-merge-patch

anguslees:
If an object does not yet exist on the server, kubecfg creates it
using a `POST` - this has not changed.

If an object already exists on the server, previously kubecfg would
`PATCH` it with a JSON merge patch to produce the updated version.
This was simple and worked for _almost all_ cases, but (eg) failed to
remove annotations that had been removed from local config.  This
latter is surprising because it is not how `kubectl apply` works, and
it is not the object-level semantics that `kubecfg` implements with
its own garbage collection.

This change switches to performing a schema-driven client-side 3-way
strategic-merge-patch between the last version of the object created
by `kubecfg`, the new version of the object found in local configs,
and the current version found on the server.  The "last version" of
the object is stored JSON-serialised in a new
`kubecfg.ksonnet.io/last-applied-configuration` annotation on each
object.  The updated object is then sent to the server using a
`PUT` (with retry on resource conflict).

This reimplements the `kubectl apply` semantics, so is consistent with
the standard kubernetes tools in this space.

Also a few minor but user-visible verbosity improvements:
- Now prints only one of "Creating ..." or "Updating ..." progress
  message.
- Prints no message at all if the object has not changed (according to
  `Semantic.DeepEqual`)

Note the `last-applied-configuration` annotation is compressed to reduce
the storage/network overhead.  For reference, the "manual" way to inspect
this annotation is:
```sh
kubectl get deployment/foo \
  -o jsonpath='{.metadata.annotations.kubecfg\.ksonnet\.io/last-applied-configuration}' |
  base64 -d | zcat | jq
```


anguslees:
Given the significance of the change, I want to test this further before we merge it...

mkmik:
Do you know how does `rc.Update` behave in case of a mutating admission controller?
In particular, if the admission controller adds back an annotation we removed during the 3-way merge, will we correctly avoid outputting "Updated"?

anguslees:
> Do you know how does rc.Update behave in case of a mutating admission controller?

I would want to test it, but my understanding is that the Update result (ie: the HTTP PUT response body) will include the results of the mutating admission controller.

Note the code does (paraphrased):
```
if DeepEqual(existing, new) {
   return nil
}
log.Info("Updating ...")
err := rc.Update(new)
```

So we print the "Updating" message every time we perform the HTTP PUT - and without regard to any changes the server might make from that point.

mkmik:
Yeah, I guess this means we should recompute the differences between the newobj (as returned by update) and the existing object (the object on the API server before our update)

anguslees:
> I guess this means we should recompute the differences between the newobj (as returned by update) and the existing object (the object on the API server before our update)

I figured/wanted the "no differences" code to kick in when kubecfg is run repeatedly with no config changes.  Specifically in this case:

- we have some initial or modified jsonnet
- rc.Create() or Update() with the orig jsonnet(A) stashed in the annotation (as json)
- (hypothetically) server or controller makes some additional (non-conflicting) changes to object, producing some result(B) on server
- kubecfg is run again with unmodified jsonnet(A')
- now we have: 3way merge of A, B, A', which should produce B (since A->A' is no changes)
- This will be detected as a no-op because state on server == B

(ie: `patch(existing, obj)` will return a copy of `existing`)

I will create some integration tests to verify this and other combinations - this is part of what I wanted to confirm before merging this PR.

mkmik:
ok, if annotations are not conflicting then the behaviour of this PR is sound and easy to reason about.

I was thinking of this scenario:

First version of jsonnet input `A` had annotation `a` and once pushed now the state `B` on server also has annotation `a`.

Now the user removes annotation `a` in `A'` , but the server reinstates the annotation `a` because the admission controller puts it back.

now, kubecfg can report this in two ways:

a) print "Updated", because in a way that's what happens: the user wanted to remove an annotation and kubecfg obliged

b) don't print "Updated" because the operation had no real world effect (the annotation is still there)

which of those two behaviours is least surprising?

(Since `A'` is then saved in "last-applied-configuration" this choice matters only once)

seh:
This is wonderful to see, @anguslees.

One question so far: Would you consider [compressing the content](https://github.com/ksonnet/kubecfg/pull/246/files#diff-fbaa30e0372d7d2671b4ba512ef44feeR188) of the "kubecfg.ksonnet.io/last-applied-configuration" annotation? Does it need to be human-readable?

anguslees:
> First version of jsonnet input A had annotation a and once pushed now the state B on server also has annotation a.
> Now the user removes annotation a in A' , but the server reinstates the annotation a because the admission controller puts it back.

(With this PR) kubecfg will see the removed annotation (A->A'), print "Updated" and update the object removing the annotation (both last-applied and actual object won't have the annotation).

If the server adds the annotation back (either admission-time or later asynchronously), then on the _next_ run kubecfg will see the annotation as added externally (A->A' does not have the annotation, B does), and patch() will preserve the annotation since it does not conflict with A->A'.  kubecfg will not print "Updated" and will not update the object (assuming there are no other A->A' changes) because the existing object matches the result of patch(A, B, A').

Obviously I need to actually do the update before I can see if/what the server might do next.  The current code just prints "Updating" (future tense) whenever it actually does an Update() call - mostly so the error makes sense in context (if the Update() fails).

I think you're suggesting I should compare the REST request vs response and then only print "Updated" (past tense) if that changed.  I can certainly do that instead (I need to strip out the server-set metadata in both cases anyway).

> Would you consider compressing the content of the "kubecfg.ksonnet.io/last-applied-configuration" annotation?

:thinking: I had not thought of compressing it, other than using a json rendering with minimal whitespace.

`kubectl apply` uses regular json, but there is no reason this needs to be human readable (or compatible - we don't _want_ to use the last-applied value for other tools).  I could compress it (eg: gzip+base64) easily enough.

A quick test on a largish object (Deployment) and smallish (Service) produced these file sizes:

 2346 | deploy.json
 1232 | deploy.json-compact
   820 | deploy.json-compact.gz.base64
   808 | svc.json
   490 | svc.json-compact
   412 | svc.json-compact.gz.base64

So it is even a win for small objects.  Good suggestion :)

anguslees:
I'm happy with the test coverage now (and uncovered/fixed several important bugs in my first iteration).
I want to do the base64+gzip thing that seh suggested before merging, however (a small change, but easier to do in the same PR so there is no migration to worry about)

davidquarles:
any update here?  I've been eagerly checking in on this PR every few days.

anguslees:
> any update here?

The increased test cases uncovered a cache-invalidation bug in this PR, which I hadn't noticed until I ran it against the empty CI test environment.  (A brand new CRD plus an instance of that custom resource in the same run failed.)

I finally fixed that this morning, yay.  Now the same test is catching a race (that was always there) between creating a CRD and when that new custom resource appears in discovery metadata.  Previously I ignored that, since it rarely triggered in practice (round trip delays to the server and updates to other intervening objects were sufficient to cover up the delay).  I figured I'd fix that while I'm here though, rather than insert a sleep() or similar hack just to get the test to pass reliably.

So: yay for tests.

anguslees:
@mkmik: I think we're good to go!  There have been a number of changes since your last review, so you probably want to look it over again.

utils/meta.go
```diff
@@ -142,3 +165,43 @@ func FqName(o metav1.Object) string {
...
+// (json->gzip->base64) of an object, intended for use in
+// last-applied-configuration annotation.
+func CompactEncodeObject(o runtime.Object) (string, error) {
+	// json.Encoder output isn't compact :(
+	var jsbuf bytes.Buffer
+	tmp, err := json.Marshal(o)
+	if err != nil {
+		return "", err
+	}
+	if err := json.Compact(&jsbuf, tmp); err != nil {
```
mkmik: Please use json.Encoder and SetIndent("","")

mkmik:
@anguslees I forgot why we didn't just use `kubectl.kubernetes.io/last-applied-configuration`. Is attempting interoperability with `kubectl apply` unsound?

PR #260, opened 2019-08-16
update: Fall back to JSON merge patch when strategic merge patch preparation fails

### LINKED ISSUE #259: "kubecfg update" fails against CRD objects: "expected kind, but got map"
# Observed Behavior

Using _kubecfg update_ against a server running Kubernetes version 1.15.1 or later, attempting to update an existing CRD object (not the CRD itself, but an object of a kind defined by a CRD), _kubecfg_ fails with the following message:
```
ERROR Error updating sealedsecrets my-ns.my-object: expected kind, but got map
```
Here we see _kubecfg_'s attempt to update an object called "my-object" in namespace "my-ns." Note that this object's source manifest hasn't changed since the last run of _kubecfg update_. This message originates from [_api-machinery_'s `(*util/strategicpatch.kindItem).VisitMap` method](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/strategicpatch/types.go#L73-L75).

We can create objects of kinds defined by a CRD, and we can delete them without any problems. However, we can't run _kubecfg update_ against such an **existing** object.

In this case, we create the CRD object itself using _kube.libsonnet_'s `CustomResourceDefinition` function:
```jsonnet
kube.CustomResourceDefinition('bitnami.com', 'v1alpha1', 'SealedSecret')
```
The corresponding YAML as shown by _kubecfg show_:
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    app: sealed-secrets
  name: sealedsecrets.bitnami.com
spec:
  group: bitnami.com
  names:
    kind: SealedSecret
    listKind: SealedSecretList
    plural: sealedsecrets
    singular: sealedsecret
  scope: Namespaced
  version: v1alpha1
```
That is, we don't define its schema, in case that matters here.

Though we use this _SealedSecret_ kind as an example here, we've experienced the same problem with similar CRDs we've defined that used to work fine with _kubecfg update_ before Kubernetes 1.15.1 (at least with 1.14.3).

Note that pulumi/pulumi-kubernetes#630 mentions this same error, but doesn't indicate how they fixed it; it sounds like the problem just went away. We have not been so lucky with our clusters.

# Software
## kubecfg version
```
kubecfg version: v0.12.4
jsonnet version: v0.12.0
client-go version: v0.0.0-master+a27bea2b
```
## Kubernetes version
```
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.1", GitCommit:"4485c6f18cee9a5d3c3b4e523bd27972b1b53892", GitTreeState:"clean", BuildDate:"2019-07-18T09:09:21Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

seh:
Kubernetes custom resources do not support strategic merge patch. Since it is difficult to discern which objects belong to custom resources, instead interpret a failure to prepare a strategic merge patch as a likely indicator that the API server does not support the patch type for the object, and fall back to trying a JSON merge patch instead.

See pulumi/pulumi-kubernetes#622 for the basis of inspiration. See kubernetes/kubernetes#81515 for a related fix that may allow us to refine our detection approach in the future.

Fixes #259

### DIFF
--- a/pkg/kubecfg/update.go
+++ b/pkg/kubecfg/update.go
@@ -120,26 +120,40 @@ func patch(existing, new *unstructured.Unstructured, schema proto.Schema) (*unst
 	}
 
 	var resData []byte
+	patchWithJSONMerge := func() error {
+		patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(origData, newData, existingData)
+		if err == nil {
+			resData, err = jsonpatch.MergePatch(existingData, patch)
+		}
+		return err
+	}
 	if schema == nil {
 		// No schema information - fallback to JSON merge patch
-		patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(origData, newData, existingData)
-		if err != nil {
-			return nil, err
-		}
-		resData, err = jsonpatch.MergePatch(existingData, patch)
-		if err != nil {
+		if err := patchWithJSONMerge(); err != nil {
 			return nil, err
 		}
 	} else {
 		patchMeta := strategicpatch.NewPatchMetaFromOpenAPI(schema)
 
 		patch, err := strategicpatch.CreateThreeWayMergePatch(origData, newData, existingData, patchMeta, true)
-		if err != nil {
-			return nil, err
-		}
-		resData, err = strategicpatch.StrategicMergePatchUsingLookupPatchMeta(existingData, patch, patchMeta)
-		if err != nil {
-			return nil, err
+		if err == nil {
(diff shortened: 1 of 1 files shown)


seh:
Note that I did not add a test for this case, not being sure what's in that _testdata/schema.pb_ file that could provoke this branch. Please let me know if you'd me to try to force such a test case.

seh:
See #261 for a more precise treatment.

PR #262, opened 2019-08-20
add diff ignore paths

theSuess:
When using kubecfg in combination with openshift, some issues arise due to openshift either having a broken api spec (e.g the status field in route objects is not optional) or updating the spec by using triggers.

This PR allows to specify one or multiple paths to ignore when calculating diffs.
For example `--ignore='spec.template.spec.containers.0.image'` ignores changes to image of the first container in a deploymentconfig.


theSuess:
i've encountered some issues with this approach. i'll rewrite this (possibly without external dependencies) and submit a new PR

PR #263, opened 2019-08-23
add flag to omit secrets in diff

theSuess:
We use the diff functionality in our CI/CD pipeline to show pending changes before each deployment. Since those pipelines can be viewed by everyone, i've added a flag to alter the diff produced by secrets. An example diff would look like this:

```.diff
---
- live secrets example
+ config secrets example
-     "SSO_TOKEN": <omitted>
+     "SSO_TOKEN": <omitted>
```

### DIFF
--- a/pkg/kubecfg/diff.go
+++ b/pkg/kubecfg/diff.go
@@ -41,7 +41,7 @@ var ErrDiffFound = fmt.Errorf("Differences found.")
 // Matches all the line starts on a diff text, which is where we put diff markers and indent
 var DiffLineStart = regexp.MustCompile("(^|\n)(.)")
 
-var DiffKeyValue = regexp.MustCompile(`"([[:alnum:]_-]+)":\s"([[:alnum:]=+]+)",?`)
+var DiffKeyValue = regexp.MustCompile(`"([-._a-zA-Z0-9]+)":\s"([[:alnum:]=+]+)",?`)
 
 // DiffCmd represents the diff subcommand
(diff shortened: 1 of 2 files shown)


pkg/kubecfg/diff.go
```diff
@@ -41,11 +41,14 @@ var ErrDiffFound = fmt.Errorf("Differences found.")
 // Matches all the line starts on a diff text, which is where we put diff markers and indent
 var DiffLineStart = regexp.MustCompile("(^|\n)(.)")
 
+var DiffKeyValue = regexp.MustCompile(`"([[:alnum:]_-]+)":\s"([[:alnum:]=+]+)",?`)
```
mkmik: > a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')

turns out `.` is also valid in a secret name; could you please update the regexp (and possibly also add a test case)?

I tried to craft such a secret name and indeed it leeks through the cracks:

```
$ ./kubecfg diff --omit-secrets secret.yaml
---
- live secrets mysecret
+ config secrets mysecret
-     "foo.x": "cXV4"
+     foo: <omitted>
```

theSuess: i've changed the regex to the one used by kubernetes internally - should now also work with keys containing `.`

mkmik:
Thanks

PR #265, opened 2019-09-22
Define "ext-code," "ext-code-file," "tla-code," and "tla-code-file" command-line flags

### LINKED ISSUE #63: Needs implementation for --ext-code and --tla-code
These were on my backburner, because I thought I didn't need them, but I ran into a situation today that required it. This issue is mostly for my benefit (I'll try to whip one out this weekend)

seh:
Mirroring the _jsonnet_ tool's command-line flags for consuming externally supplied Jsonnet code as injected values, define the four new flags:
- `--ext-code`
- `--ext-code-file`
- `--tla-code`
- `--tla-code-file`

Fixes #63.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -48,15 +48,19 @@ import (
 )
 
 const (
-	flagVerbose    = "verbose"
-	flagJpath      = "jpath"
-	flagJUrl       = "jurl"
-	flagExtVar     = "ext-str"
-	flagExtVarFile = "ext-str-file"
-	flagTlaVar     = "tla-str"
-	flagTlaVarFile = "tla-str-file"
-	flagResolver   = "resolve-images"
-	flagResolvFail = "resolve-images-error"
+	flagVerbose     = "verbose"
+	flagJpath       = "jpath"
+	flagJUrl        = "jurl"
+	flagExtVar      = "ext-str"
+	flagExtVarFile  = "ext-str-file"
+	flagExtCode     = "ext-code"
+	flagExtCodeFile = "ext-code-file"
+	flagTLAVar      = "tla-str"
+	flagTLAVarFile  = "tla-str-file"
+	flagTLACode     = "tla-code"
+	flagTLACodeFile = "tla-code-file"
+	flagResolver    = "resolve-images"
+	flagResolvFail  = "resolve-images-error"
 )
 
 var clientConfig clientcmd.ClientConfig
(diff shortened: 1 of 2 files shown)


seh:
The build failed due to the following fetch attempt failing:
> fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out

Is that a known problem? How we're depending on Thrift is beyond me.

cmd/root.go
```diff
@@ -275,6 +285,41 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 		vm.ExtVar(kv[0], string(v))
 	}
 
+	extcodes, err := flags.GetStringSlice(flagExtCode)
+	if err != nil {
```
mkmik: The basic code for dealing with variables is repeated for every flavour of variables.

Since the method signatures of vm.ExtVar, vm.ExtCode, VM.TLAVar vm.TLACode are the same, couldn't we factor this code out into a helper function that has a func(string,string) parameter (In Go you can pass a bound method where a function is expected)

mkmik:
We should be using vendoring or a module proxy.
It seems that we call go generate without passing -mod=vendor

cmd/root.go
seh: Hence this comment in my original description:
> The implementation adapts the existing code for the `--ext-str`, `--ext-str-file`, `--tla-str`, and `--tla-str-file flags`, without factoring parts that could possibly be shared.

That was my way of saying that as a first pass, I didn't add anything novel, to probe whether you would prefer making the individual treatment of these eight flags more similar.

To that end, please see two different styles of doing so:
- Commit 59df316a9b79ed9bf2b0dd1eef821d8cc536ae22 introduces two helper functions, and calls on them eight times.
- Commit 50a0249020be669cccffe2ae42a2231db504f692 replaces those functions with iteration over a table, dispatching on the string-versus-file variation.

I prefer the latter. If there's a different style you'd prefer, please let me know.

seh:
> It seems that we call go generate without passing -mod=vendor

Do you want me to try fixing that problem as part of this patch? Perhaps adding `$(GO_FLAGS)` on line 43 of the _Makefile_ would be sufficient:
```make
generate:
	$(GO) generate -x $(GO_FLAGS) $(GO_PACKAGES)
```
It looks like _go generate_ invokes _go-bindata_. It's not clear which piece of this chain is responsible for fetching all of these dependencies, and which piece would wind up caring about the `-mod=vendor` flag.

cmd/root.go
```diff
@@ -240,74 +250,47 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	for _, extvar := range extvarfiles {
-		kv := strings.SplitN(extvar, "=", 2)
-		if len(kv) != 2 {
-			return nil, fmt.Errorf("Failed to parse %s: missing '=' in %s", flagExtVarFile, extvar)
-		}
-		v, err := ioutil.ReadFile(kv[1])
+	for flagName, spec := range map[string]struct {
```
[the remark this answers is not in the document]
seh: Please see commit bbb3ab64070c723ad555349ebcc8b916a42fcb05 for the conversion from a map to a slice.

mkmik:
W.r.t the thrift issue, I'm ok with fixing it in this PR or in a separate one, as you prefer.

I don't know which part of the build triggers the module fetch.

Unfortunately I'm AFK until the evening.

seh:
> W.r.t the thrift issue, I'm ok with fixing it in this PR or in a separate one, as you prefer.

Reading about similar trouble that's plagued many other Go projects, it sounds like we pick up Thrift through the Go project's reliance on OpenCensus.

I added `-mod=vendor` to the _go generate_ invocation by including the `GO_FLAGS` variable in that _Makefile_ rule. I'm not sure that `-mod=vendor` belongs in `GO_FLAGS`, as opposed to `GO_BUILDFLAGS`, given that it has influence over commands that build Go packages.

In any case, commit dc01d593294c0ab2ad7f55ca5e077c4b735b7d85 appears to have appeased Travis CI for now.

Once we merge this patch, I will be eager for a release that includes it.

mkmik:
Thanks!

Yeah I think a release is a good idea.

PR #271, opened 2019-10-06
Detect imported extvars and TLAs via sentinel import path

seh:
When a Jsonnet consults an Importer for a path nominated by an external variable or top-level argument file-based command-line flag, the "importedFrom" parameter passed to [the `Importer.Import` method](https://github.com/google/go-jsonnet/blob/master/imports.go#L29-L51) is not a valid URL, but is instead a sentinel placeholder value—namely,
an empty string. Detect these sentinel values and use the current working directory as the base URL instead when resolving relative import paths.

**NB:** This patch is speculative for now, based on a change to the Jsonnet `Importer.Import` method's contract proposed by @sbarzowski (https://github.com/bitnami/kubecfg/pull/266#issuecomment-535090890).


seh:
See google/go-jsonnet#447. With that, we can proceed here. @mkmik, take note.

PR #272, opened 2019-10-15
validate if name is set

theSuess:
This change simply checks if the name of the object is set. Without this check, the validation succeeds on typos like this:
```jsonnet
    metadata: { labels+: {"foo":"bar"}}
```
but will panic while updating with the following error, making it especially hard to debug:
```
panic: interface conversion: runtime.Object is *unstructured.UnstructuredList, not *unstructured.Unstructured
```

### DIFF
--- a/pkg/kubecfg/validate.go
+++ b/pkg/kubecfg/validate.go
@@ -82,6 +82,9 @@ func (c ValidateCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer)
 			for _, err := range schema.Validate(obj) {
 				allErrs = append(allErrs, err)
 			}
+			if obj.GetName() == "" {
+				allErrs = append(allErrs, fmt.Errorf("An Object does not have a name set"))
+			}
 		}
 
 		for _, err := range allErrs {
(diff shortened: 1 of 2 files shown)


pkg/kubecfg/validate.go
```diff
@@ -82,6 +82,9 @@ func (c ValidateCmd) Run(apiObjects []*unstructured.Unstructured, out io.Writer)
 			for _, err := range schema.Validate(obj) {
 				allErrs = append(allErrs, err)
 			}
+			if obj.GetName() == "" {
```
seh: What about [`ObjectMeta.GenerateName`](https://github.com/kubernetes/apimachinery/blob/4b0ceda2f81e6ad8a403818599068eaed29e1724/pkg/apis/meta/v1/types.go#L120-L136)? Can _kubecfg_ accommodate its use?

theSuess: We could generate a name ourselves using this field if it exists but server-side name generation would not work because of the way we fetch the existing resource before applying the config (it would also not support diffing).

mkmik: yeah, generateName doesn't seem compatible with the workflow promoted by kubecfg

PR #279, opened 2019-12-18
Changes to be able to use 1Password to store values from secrets and retrieve them from 1Password

thomasmhofmann:
These changes integrate kubecfg with 1Password.

The rational behind these changes is the following:
* Passwords for different environments (dev, test, prod) should be different and it should be possible to generate good passwords automatically.
* These generated passwords need to be stored somewhere so that a user can look them up.
* They should not be stored in SCM.
* Passwords / secrets need to be shared among team members.

The proposed solution add some basic / general new native functions to kubecfg.libsonnet. Using these primitives a Jsonnet library (1password.libsonnet) is built that provides 1Password-specific functionality.

To use the integration one would create a 'OnePasswordSecret' instead of a 'Secret' (from kube.libsonnet).
This special secret will determine the name of the vault to use in 1Password and will also determine a name for the new item where the values from the secret should be stored.
It will then try to read that item from 1Password and incorporate the information found into the stringData field of the kubernetes secret resource.
If the item does not exist, it will create a new item and store the information in this new item to the 1Password vault. Incremental changes are not supported. So when the secret definition in Jsonnet is changed (a new field is added) it will not be stored to 1Password.

The PR includes a sample secret kubernetes resource in Jsonnet and the 1Password library.

Some of the things that is included in the 1Password library could have also been added directly in kubecfg but I did not want too speciifc things to kubecfg. Unfortunately, I still needed to add four new functions to kubecfg.libsonnet (actually one is not really needed for the integration).

This is my first PR and I hope that I did not do too many mistakes. I used different commits for changes to vendor.


thomasmhofmann:
Hi there, any chance to get this merged? Or should I take another approach? Can we discuss what would be an acceptable change?

mkmik:
I'll review soon, sorry you caught me in a busy moment (paternity)

thomasmhofmann:
Congratulations, I just did not see any reaction that's why I decided to ask now. Take your time.

Issue #280, opened 2020-01-22
ext-code/tla-code improperly parsed


camh-:
The command line flags `--ext-code` and `--tla-code` do not handle values with commas or double quotes in them and produce an error when you try:

    $ kubecfg show --ext-code 'foo={a: 1, b: 2}'
    ERROR Missing environment variable:  b: 2}
    
    $ kubecfg show --ext-code 'foo="bar"`
    ERROR invalid argument "foo=\"bar\"" for "--ext-code" flag: parse error on line 1, column 4: bare " in non-quoted-field

This happens because the flag is parsed with Cobra's [`StringSlice()`](https://godoc.org/github.com/spf13/pflag#FlagSet.StringSlice) flag type which supports multiple comma separated values. This explains why the comma doesn't work. The double quotes do not work because `StringSlice()` parses values with `encoding/csv` - i.e. as a CSV value. This escapes double quotes.

It does not look like there is any way to escape the value such that it can be parsed as intended.

I think the correct fix is to parse these flags with [`StringArray()`](https://godoc.org/github.com/spf13/pflag#FlagSet.StringArray) instead which does not try to split the value on commas (and no CSV parsing to muck up double quotes). I think this change should be made for all 8 variants of ext/tla options so that commas and double-quotes can be used in string values and filenames. This is also consistent with the `jsonnet` cli, which does not support comma-separated flag values.

PR #281, opened 2020-01-22
Parse tla/extvar options as StringArray not StringSlice

camh-:
The command line flags for top-level args (TLAs) and external variables
(extVars) were being parsed with the Cobra flag type of StringSlice. A
StringSlice allows multiple arguments with one flag name by separating
the values with a comma. This means a comma cannot be used in the
values. Also, Cobra StringSlice parses the argument with encoding/csv
which escapes double-quotes in the value.

For code arguments, we need the value to be passed unchanged. Cobra
StringArray does this, but does not allow the form --var=1,2,3 to provide
a list of [1, 2, 3].

Change all of the --ext and --tla argument processing to use StringArray
instead of StringSlice. This will change the behaviour for parsing
non-code argument, but it could be argued that using StringSlice in the
non-code arguments is also wrong. --ext-str foo=bar,baz means set foo to
bar and baz to the value from the environment. There is no way to set
foo to "bar,baz".

So let's make all cases possible at the expense of slightly more
verbosity in a few cases (--ext-str foo=bar --ext-str baz) and use
StringArray everywhere. This is also in line with the jsonnet command
line parsing.

Update the test cases to include a test for an extVar with a comma and
double quotes.

Fixes: #280


mkmik:
Thanks.

PR #287, opened 2020-04-09
feat: update to v1.17 apis

jjo:
* k8s deps updated to @v0.17.4, needed for projects using
  kubecfg as dependency and using >= 1.16 kube API
* googleapis/gnostic pinned to v0.4.0, see go.mod notes
* utils/client.go updated for newer discovery api
* updated integration test suite


mkmik:
```
integration/integration_suite_test.go:100:9: undefined: unstructured.
```

mkmik:
The integration tests are compiled with a build tag (iirc --tags integration) so you might not have noticed some tests used deprecated api

jjo:
@mkmik thanks for the feedback, updated the integration test suite, including bumping minikube (to latest) y Kube version (to latest v1.16)

PR #300, opened 2021-05-04
Upgrade to go-jsonnet v0.16.0

mkmik:
v0.17.0 has some backward compatibility issues. I'd like to first bump to go-jsonnet v0.16.0 and then to v0.17.0


absoludity (approved):
Do we still need to be vendoring the mods?

mkmik:
> Do we still need to be vendoring the mods?

technically no; we could remove all that crap (in another PR)

aruiz14 (approved):

PR #303, opened 2021-05-08
add last-applied strategy for diff

gaurav517:

PR #304, opened 2021-05-10
Support kubecfg.ksonnet.io/last-applied-configuration annotation in diff

mkmik:
in #303 @gaurav517 added the awesome `--diff-strategy=last-applied` feature.

This PR adds support for the kubecfg's native `kubecfg.ksonnet.io/last-applied-configuration` (We added that annotation before kubectl had one IIRC; we should probably start using the official kubectl annotation, but we must also support the old kubecfg one!)


aruiz14 (approved):
LGTM!

PR #309, opened 2021-06-05
implement local file cache for HTTP URLs

tinyzimmer:
This PR implements a dumb local cache for files fetched remotely. Cache location defaults to `<user_cache_dir>/kubecfg`. Files are written to subdirectories matching the "path" fields in the URL. 

The location of the cache is configurable with the `--cache-dir` flag.

### DIFF
--- a/utils/cache.go
+++ b/utils/cache.go
@@ -1,0 +1,154 @@
+package utils
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	assetfs "github.com/elazarl/go-bindata-assetfs"
+	jsonnet "github.com/google/go-jsonnet"
+	log "github.com/sirupsen/logrus"
+)
+
+var errNotFound = errors.New("Not found")
+
+// cache implements a dumb local cache for files fetched remotely.
+type httpCache struct {
+	// The location of the cache directory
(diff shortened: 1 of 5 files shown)


utils/cache.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return &httpCache{
+		cacheDir: cacheDir,
+		httpClient: &http.Client{
+			Transport: t,
+		},
+	}
+}
+
+var httpRegex = regexp.MustCompile("^(https?)://")
```
tinyzimmer, line 57: I was originally going to just escape the `//` - hence the capture group.

Ultimately, liked just stripping the protocol and nesting directories more. Makes for an easier to manage cache as well imo.

utils/cache.go
```diff
@@ -0,0 +1,154 @@
...
+			return err
+		}
+		if err := os.MkdirAll(localPathDir, 0755); err != nil {
+			return err
+		}
+	}
+	if err == nil && !finfo.IsDir() {
+		return fmt.Errorf("%q is not a directory, it cannot be used for caching", localPathDir)
+	}
+	return ioutil.WriteFile(localPath, contents, 0644)
```
tinyzimmer, line 88: I don't know off hand if this will replace an existing symlink to nothing (the afore discussed power-user way of disabling caching for certain things) or return an error. I should test it probably before this is merged. ​I assume if it doesn't error, we'll want to check for a pre-existing dead link or whatever to make sure we don't mess with something the user did on purpose.

PR #310, opened 2021-07-06
Introduce new diff flag --ignore-status-field

kgtw:
Introduce new diff flag `--ignore-status-field` which can be used to ignore the top-level `.status` field when comparing diffs.

In some cases when performing a diff on resources the only shown change is the `.status` field. Seeing as this field is only updated on the live object, it doesn't make much sense to show a diff of it.

### DIFF
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -22,13 +22,15 @@ import (
 )
 
 const (
-	flagDiffStrategy = "diff-strategy"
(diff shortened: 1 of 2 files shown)


mkmik:
would `--diff-strategy last-applied` work for you?

kgtw:
Yes, `--diff-strategy last-applied` works. We were previously using `subset`.

